PR #282, opened 2014-12-19
Updated readme with execution example.  Also typos, etc.

rowandavies:


stephanh:
@rowandavies where is this up to?

README.md
```diff
@@ -76,172 +76,58 @@ This is a simplified example, a real data set may have 100s of
 columns, but it should be enough to demonstrate. The important points
 here are that _order_ is important, the struct should be defined to
 have fields in the same order as input data, and _types_ are
-important, they should accurately descript the data (and will be used
+important, they should accurately describe the data (and will be used
 to infer how the data should be parsed and validated).
 
 
-### Building a pipeline from the built in tools
+### Building an `Execution` pipeline from the built in tools
```
SamRoberts: I don't know if the word `pipeline` is the best one to use with the new execution API. It makes a lot of sense when you are thinking in terms of explicit data dependencies (this flows into that), but I'm not so sure it suits a `for-yield` comprehension.

README.md
```diff
@@ -76,172 +76,58 @@ This is a simplified example, a real data set may have 100s of
...
-can be implemented with _just_ `maestro`, and the corner cases can
-be easily handled by hive or raw scalding jobs.
+A pipeline is defined via a scalding `Execution`, generally involving
+multiple steps, with each step potentially depending on the results of
+previous steps.  This is often neatly expressed as a Scala
+`for`-`yield` comprehension like in the example below.
 
-A pipeline built only from `maestro` tasks.
+The steps in a pipeline can include scalding jobs, hive queries,
+sqoop import/export, hdfs operations, other maestro `Executions` and
```
SamRoberts: minor: would prefer "`Execution`s" to "`Executions`".

README.md
```diff
@@ -76,172 +76,58 @@ This is a simplified example, a real data set may have 100s of
...
-### Building a pipeline from the built in tools
+### Building an `Execution` pipeline from the built in tools
 
-A pipeline is defined in terms of a `cascade`. This terminology comes
-from the underlying technology used, but it is easier to think of it
-as a set of jobs (with an implied partial-ordering, based on data
-dependencies).
 
-A cascade can be made up of scalding jobs, hive queries (future),
-shell commands, or `maestro` tasks. The hope is that most pipelines
```
SamRoberts: Personally, I feel that this sentence about most pipelines just using out of the box maestro functionality, but corner cases handled via scalding / hdfs / hive commands, is quite important. It communicates the intent of the API, which I think is nice in a readme.

README.md
```diff
@@ -76,172 +76,58 @@ This is a simplified example, a real data set may have 100s of
...
-can be implemented with _just_ `maestro`, and the corner cases can
-be easily handled by hive or raw scalding jobs.
+A pipeline is defined via a scalding `Execution`, generally involving
+multiple steps, with each step potentially depending on the results of
+previous steps.  This is often neatly expressed as a Scala
+`for`-`yield` comprehension like in the example below.
 
-A pipeline built only from `maestro` tasks.
+The steps in a pipeline can include scalding jobs, hive queries,
```
SamRoberts: we should probably list out-of-the-box maestro functionality _first_, followed by custom functionality like dropping down to scalding and hive.

On the other hand, I guess at this point in the readme, the users haven't been introduced to the various maestro concepts yet, so that's not going to be so easy! How can we sum up the out-of-the-box maestro functionality? Something like "out-of-the-box maestro executions to import data, apply strongly-typed schemas, combine, partition, and write data into various formats"?

README.md
```diff
@@ -76,172 +76,58 @@ This is a simplified example, a real data set may have 100s of
...
-    Clean.trim,
-    Clean.removeNonPrintables
+/** Configuration for a customer execution example */
+case class CustomerConfig(config: Config)  {
+  val maestro   = MaestroConfig(/*...*/)
+  val upload    = maestro.upload()
+  val load      = maestro.load[Customer](none="null")
+  val dateTable = maestro.partitionedHiveTable[Customer, (String, String, String)] (/*...*/)
+  val catTable  = maestro.partitionedHiveTable[Customer, String](/*...*/)
+  val queries = List(
```
SamRoberts: These queries duplicate work done by the view statements, so it's not really the best code. We just wanted to show off different features. I am happy with this for now, but I am going to try to change the `CustomerExecution` example to be more sensible (at the expense of not showing off all features), and have another more contrived example to show off all the features.

README.md
```diff
@@ -253,12 +139,34 @@ to generate out significant supporting infrastructure customized for
 _your_ specific record type. This gives us the ability to refer to fields
 for partitioning, filtering and validation in a way that can be easily
 be checked and validated up front (by the compiler in most circumstances,
-and on start-up before things run in the worst case) and that those field
+and on start-up before things run in the worst case).  Such field
```
SamRoberts: @stephanh what do you think about this sentence: "Such field references can have large amounts of interesting metadata which allows us to [do lots of stuff automatically]". I know we have vague plans to add more metadata to fields, but right now we are talking about a thrift type, and we don't get automatic validation, filtering, or partitioning. IMO this sentence badly oversells maestro's current capabilities.

README.md
```diff
@@ -253,12 +139,34 @@ to generate out significant supporting infrastructure customized for
...
+also providers that are allowed to perform different forms of effects prior
+to providing an item like throwing exceptions (`Try[T]`) 
+changing state (`scalaz.effect.ST[T]`) and arbitrary operations including I/O (`scalaz.effect.IO[T]`).  
+
+Using `Execution` doesn't require a deep understanding
+of monads, mostly it just requires chaining some basic executions 
+together to form sequences.
+
+TODO Unfinished - roughly explain the sense in which `Execution[T]` provides items.
+TODO Add references for more info, maybe say less about general monads.
```
SamRoberts: Agree with this TODO. I don't think we should attempt an explanation of monads in this section. Instead, just explain how you can compose Executions with the `for-yield` notation to produce another Execution. (I wouldn't even go into flatMap personally, but this is debatable).

README.md
```diff
@@ -336,6 +244,67 @@ Validator.all(
 | double: A 64-bit floating point number            | DOUBLE (8-byte double precision floating point number)                                        | double        |
 | string: Encoding agnostic text or binary string   | string 
 
+### Hive
```
SamRoberts: We probably want to discuss the Hive monad and (separately) the Hdfs monad, and show how we allow them to be used within an Execution (maybe with a small example).

README.md
```diff
@@ -336,6 +244,67 @@ Validator.all(
...
+Maestro allows you to write directly to Hive tables and to run queries on Hive tables as part of a
+Maestro job.
+
+Create a `HiveTable` to describe/reference a specific hive table. This is required by the other hive
+related methods as identifier. The `source` and `sink` methods on the `HiveTable` provide Scalding
+sources and sinks for typed pipes to read from or write to. `name` provides a fully qualifed name
+that can be used inside hql.
+
+`viewHive` allows the Maestro job to write out the data to a partitioned hive table in parquet
+similar to `view`. However, it also creates the hive table if it doesn't already exist. Otherwise,
```
SamRoberts: Personally, I think the last sentence ("Otherwise, it just ..."), while true, is probably not necessary.

README.md
```diff
@@ -336,6 +244,67 @@ Validator.all(
...
+it just verifies the schema.
+
+`hiveQuery` enables the running of a hive query as part of the Maestro job. Apart from the query
```
SamRoberts: We have replaced `hiveQuery` with the `Hive` monad

SamRoberts:
@rowandavies thanks, I think this is starting to look much better. It's a big job. I assume there are still more sections you intend to change?

README.md
```diff
@@ -76,85 +78,194 @@ This is a simplified example, a real data set may have 100s of
...
+object CustomerJob extends MaestroJob {
+  /** Create an example customer execution */
+  def job: Execution[JobStatus] = for {
+    conf           <- Execution.getConfig.map(CustomerConfig(_))
+    uploadInfo     <- upload(conf.upload)
+    sources        <- uploadInfo.withSources
+    (pipe, ldInfo) <- load[Customer](conf.load, uploadInfo.files)
+    loadSuccess    <- ldInfo.withSuccess
+    (count1, _)    <- viewHive(conf.dateTable, pipe) zip viewHive(conf.catTable, pipe)
+    _              <- Execution.fromHive(Hive.queries(conf.queries), /*...*/)
```
SamRoberts: I suggest just getting rid of the `/*...*/`

README.md
```diff
@@ -76,85 +78,194 @@ This is a simplified example, a real data set may have 100s of
...
+  } yield (JobFinished)
+  /*...*/
```
SamRoberts: and getting rid of it here too

README.md
```diff
@@ -76,85 +78,194 @@ This is a simplified example, a real data set may have 100s of
...
+}
 ```
+(TODO This should be updated to be consistent with issue #295, including the yield.)
```
SamRoberts: I'll take care of this TODO, don't worry about it for your PR

README.md
```diff
@@ -76,85 +78,194 @@ This is a simplified example, a real data set may have 100s of
...
+
+An `Execution[T]` represents some work that can be performed that
+provides an item of type `T` if it succeeds, otherwise it fails.  The
+work can involve hadoop map-reduce jobs via scalding, including counters, and caching
+of results is performed when appropriate.  Many `Execution`s can be chained
+together using a `for`-`yield`-comprehension (ultimately via `flatMap`,
+see this [FAQ](http://docs.scala-lang.org/tutorials/FAQ/yield.html)) like in
+the example above.
+
+The type constructor `Execution[T]` is a monad, which roughly means that
```
SamRoberts: I believe we are better off removing this paragraph. I don't think we can take on the job of explaining what a monad is in a single paragraph. If we must mention monads, I would suggest pointing to some reference material, and than pointing to our own examples to show how it works "in practice".

stephanh:
What is the status of this PR?

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
-    maestro.ivory(clean, features),
-    maestro.sqoop(clean, output, filters)
+```scala
+/** Configuration for a customer execution example */
+case class CustomerConfig(config: Config)  {
+  val maestro   = MaestroConfig(...)
+  val upload    = maestro.upload()
+  val load      = maestro.load[Customer](none="null")
+  val dateTable = maestro.partitionedHiveTable[Customer, (String, String, String)] (...)
+  val catTable  = maestro.partitionedHiveTable[Customer, String](...)
```
SamRoberts: I would get rid of `catTable`, `queries`, and the corresponding parts of the `CustomerJob` below, personally.

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
   )
 }
 
+/** Customer execution example */
+object CustomerJob extends MaestroJob {
+  def job: Execution[JobStatus] = for {
+    conf           <- Execution.getConfig.map(CustomerConfig(_))
+    uploadInfo     <- upload(conf.upload)
+    sources        <- uploadInfo.withSources
+    (pipe, ldInfo) <- load[Customer](conf.load, uploadInfo.files)
```
SamRoberts: can use `sources` here instead of `uploadInfo.files`

SamRoberts: maybe use `loadInfo` instead of `ldInfo`? I feel bad about doing this shorter version in our examples, but I think it's especially good to use the full words here, as it's the first example of maestro code people will see.

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
+    loadSuccess    <- ldInfo.withSuccess
```
SamRoberts: we aren't using `loadSuccess` here, so you could replace this call with `if ldInfo.continue` or `_ <- ldInfo.withSuccess`

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
+    (count1, _)    <- viewHive(conf.dateTable, pipe) zip viewHive(conf.catTable, pipe)
```
SamRoberts: not using `count1`, so can probably just use `_` instead

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
+| i16: A 16-bit signed integer                      | SMALLINT (2-byte signed integer, from -32,768 to 32,767)                                      | short         |
+| i32: A 32-bit signed integer                      | INT (4-byte signed integer, from -2,147,483,648 to 2,147,483,647)                             | int           |
+| i64: A 64-bit signed integer                      | BIGINT (8-byte signed integer, from -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807)  | BigInteger    |
+| double: A 64-bit floating point number            | DOUBLE (8-byte double precision floating point number)                                        | double        |
+| string: Encoding agnostic text or binary string   | string 
+
+### Execution monad
+(See the scalding [`Execution` trait](http://twitter.github.io/scalding/#com.twitter.scalding.Execution) and [`Execution` object](http://twitter.github.io/scalding/#com.twitter.scalding.Execution$)
+as well as the maestro extensions in [`RichExecution`](https://commbank.github.io/maestro/latest/api/index.html#au.com.cba.omnia.maestro.core.scalding.RichExecution))
```
SamRoberts: Link to the RichExecution object as well?

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
+
+An execution is an object with type `Execution[T]` representing some
+work that can be performed that provides an item
+of type `T` if it succeeds, otherwise it fails.  The work can depend
+on configuration information, and can involve a variety of smaller
+steps, includinghadoop jobs via scalding, with counters and caching
```
SamRoberts: missing space between "including" and "hadoop"

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
+[`onException`](https://commbank.github.io/maestro/latest/api/index.html#au.com.cba.omnia.maestro.core.scalding.RichExecution).
+
+### Custom executions
+
+`Execution.from` allows any Scala code to be included in an `Execution`, however this should be done
+carefully, including considering handling errors and other unusual situations.  If the code
+throws an exception this is caught and converted into a failing `Execution`.
+
+Pure expressions that just
+return a value and never throw exceptions or perform effects should be included in comprehensions
```
SamRoberts: effects -> side effects?

README.md
```diff
@@ -76,73 +88,144 @@ This is a simplified example, a real data set may have 100s of
...
+
+### Hive
 
 Maestro allows you to write directly to Hive tables and to run queries on Hive tables as part of a
 Maestro job.
 
-Create a `HiveTable` to describe/reference a specific hive table. This is required by the other hive
-related methods as identifier. The `source` and `sink` methods on the `HiveTable` provide Scalding
+`HiveTable` instances allow you to describe/reference a specific hive table.
+The `source` and `sink` methods on the `HiveTable` provide Scalding
```
SamRoberts: Can also use `viewHive`

README.md
```diff
@@ -275,38 +289,21 @@ partition a data set by.
 The primary api is the list of fields you want to partition on:
 
 ```scala
-Partiton.byFields(Fields.CUSTOMER_CAT, Fields.CUSTOMER_SUB_CAT)
+Partition.byFields(Fields.CUSTOMER_CAT, Fields.CUSTOMER_SUB_CAT)
 ```
 
-The api also has special support for dates of the `yyyy-MM-dd` form:
+The api also has support for custom date formats, and
+special support for dates of the `yyyy-MM-dd` form:
```
SamRoberts: This isn't really true. The old "support" for `yyyy-MM-dd` was actually a hack that didn't behave like a standard `yyyy-MM-dd` date time pattern would. Because of that, the default format if the user does not supply one is actually `y-M-d`.

However, I wouldn't be encouraging people to use `y-M-d`. I would just avoid all mention of a special format and provide a format in the call below: `Partition.byDate(Fields.EFFECTIVE_DATE, "yyyy-MM-dd")`

README.md
```diff
@@ -324,22 +321,6 @@ Validator.all(
...
-
-### Advanced tips & tricks
+### Row filters, cleaners, splitters, transformers, macros
 
-The best tip for advanced pipelines is to look carefully at how
-the maestro tasks are implemented. You have the same set of tools
-available to you and can jump down to the same lower-level of
-abstraction by just implementing a scalding job, and extending the
-`Maestro` class.
+[TBD]
```
SamRoberts: TDB? Can we move this into an issue somewhere?

SamRoberts:
Just a few small fixes away from being done. :smile:

README.md
```diff
@@ -76,98 +88,156 @@ This is a simplified example, a real data set may have 100s of
...
+/** Customer file load job with execution for the main program */
+object CustomerJob extends MaestroJob {
+  def job: Execution[JobStatus] = for {
+    conf             <- Execution.getConfig.map(CustomerJobConfig(_))
+    uploadInfo       <- upload(conf.upload)
+    sources          <- uploadInfo.withSources
+    (pipe, loadInfo) <- load[Customer](conf.load, uploadInfo.files)
+    if loadInfo.continue
+    count            <- viewHive(conf.dateTable, pipe)
+    ()                = logger.info(s"Loaded record count: ${count}")
```
SamRoberts: Side effecting functions should go in `Execution.from` :smile:

SamRoberts:
Looks good, just need to fix the `logger.info` statement in `CustomerJob`, and :+1:

SamRoberts:
:+1:

PR #288, opened 2015-01-06
Hive monad

### LINKED ISSUE #250: Create Hive table with text format from Thrift struct

stephanh:
Takes advantage of the Hive monad from Ebenezer and all corresponding changes.

* `Result` is now in omnitool instead of permafrost.
* Use `fold` on `Result` instead of pattern matching.
* `fromHive` for Execution.
* Deprecated `QueryExecution`.
* Added Hive and Hdfs to api package.

This also closes #250.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/exec/QueryExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/exec/QueryExecution.scala
@@ -31,6 +31,7 @@ import au.com.cba.omnia.maestro.core.hive.HiveTable
   * @param queries:  The hive queries to run
   * @param settings: Hive settings. Empty by default.
   */
+@deprecated("Use the query/queries method in Hive instead", "1.15")
 case class QueryConfig(
   name: String,
   queries: Seq[String],
@@ -38,8 +39,10 @@ case class QueryConfig(
 )
 
 /** Factory functions for query configs */
+@deprecated("Use the query/queries method in Hive instead", "1.15")
 object QueryConfig {
   /** QueryConfig with empty hive settings */
+  @deprecated("Use the query/queries method in Hive instead", "1.15")
   def apply(
     name: String,
     queries: String*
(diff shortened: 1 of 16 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/exec/QueryExecution.scala
```diff
@@ -31,6 +31,7 @@ import au.com.cba.omnia.maestro.core.hive.HiveTable
   * @param queries:  The hive queries to run
   * @param settings: Hive settings. Empty by default.
   */
+@deprecated("Use the query/queries method in Hive instead", "1.15")
```
SamRoberts: :+1: for making use of deprecated tags. Should we ensure we have a github issue to remove these methods at a later date?

stephanh: Added an issue.

maestro-example/src/main/scala/au/com/cba/omnia/maestro/example/CustomerUploadExample.scala
```diff
@@ -35,15 +33,21 @@ class CustomerUploadExample(args: Args) extends Maestro[Customer](args) {
...
-    case Error(_) => {
+  List(byDateResult, byIdResult).sequence_.foldAll(
+    result => {
+      // continue with post-processing here
+    },
+    errorMsg => {
+      // do not post-process files when there was an error copying them
+      // report the error here
+    },
+    execption => {
```
SamRoberts: execption -> exception

SamRoberts:
:+1:

maestro-example/src/main/scala/au/com/cba/omnia/maestro/example/CustomerUploadExample.scala
stephanh: Fixed

PR #289, opened 2015-01-06
Minor fix to handle missing tuplesWritten count in SqoopExecution.archive.

rowandavies:


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/exec/SqoopExecution.scala
```diff
@@ -280,7 +280,7 @@ object SqoopEx extends Sqoop {
         TypedPipe.from(TextLine(src))
           .writeExecution(CompressibleTypedTsv[String](dest))
           .getAndResetCounters
-          .map(_._2.get(StatKeys.tuplesWritten).get)
+          .map( _._2.get(StatKeys.tuplesWritten).getOrElse(0L) )
```
SamRoberts: we don't generally put spaces between function call brackets and their arguments

SamRoberts:
nothing minor about that fix :)

The travis test for the commit failed. The failing test in question dies fail spuriously from time to time, due to a concurrency bug in one of the libraries we use. Still, it would be good to re-run the travis test to be sure.

Apart from that, and the minor syntactic comment above: :+1:

rowandavies:
Travis test passed.  I fixed the spaces, but messed up and committed without them, so they ended up being a separate commit:  9e1fbc21232889bd66cf0989e39b33c6f8b828fc

rowandavies:
Note that #281 seems similar.

PR #291, opened 2015-01-07
Adds append for unpartitioned hive tables using the execution api.

stephanh:
Adds the ability to append to an unpartitioned hive table. Since Hadoop doesn't allow for writing to a directory that already exists we instead write to a temporary folder, munge the file names and then copy them across.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
@@ -14,6 +14,8 @@
 
 package au.com.cba.omnia.maestro.core.hive
 
+import scalaz._, Scalaz._
+
 import org.apache.hadoop.fs.Path
 
 import org.apache.hadoop.hive.conf.HiveConf
@@ -21,10 +23,13 @@ import org.apache.hadoop.hive.conf.HiveConf.ConfVars.METASTOREWAREHOUSE
 
 import cascading.flow.FlowDef
 
-import com.twitter.scalding._
+import com.twitter.scalding.{Hdfs => _, _}
 
 import com.twitter.scrooge.ThriftStruct
 
+import au.com.cba.omnia.permafrost.hdfs.Hdfs
+
+import au.com.cba.omnia.ebenezer.scrooge.ParquetScroogeSource
 import au.com.cba.omnia.ebenezer.scrooge.hive._
 
 import au.com.cba.omnia.maestro.core.partition.Partition
@@ -100,18 +105,51 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
 ) extends HiveTable[A, A] {
 
   override def source =
-    new HiveParquetScroogeSource[A](database, table, externalPath)
+    HiveParquetScroogeSource[A](database, table, externalPath)
 
   override def sink(append: Boolean = true) =
-    new HiveParquetScroogeSource[A](database, table, externalPath)
+    if (append) throw new Exception("APPEND IS CURRENTLY NOT SUPPORTED")
+    else        HiveParquetScroogeSource[A](database, table, externalPath)
 
-  /** Writes the contents of the pipe to this HiveTable. NB: APPEND IS CURRENTLY NOT SUPPORTED. */
+  /** Writes the contents of the pipe to this HiveTable. */
   override def writeExecution(pipe: TypedPipe[A], append: Boolean = true): Execution[ExecutionCounters] = {
-    if (append) System.err.println("APPEND IS CURRENTLY NOT SUPPORTED")
-    pipe
-      .writeExecution(sink(append))
+    // Creates the hive table and gets its path
+    val setup = Execution.fromHive(
+      Hive.createParquetTable[A](database, table, List.empty, externalPath.map(new Path(_))) >>
+      Hive.getPath(database, table)
+    )
+
+    // Runs the scalding job and gets the counters
+    def write(path: Path) = pipe
+      .writeExecution(ParquetScroogeSource[A](path.toString))
       .getAndResetCounters
       .map(_._2)
+
+    //If we are appending the files are written to a temporary location and then copied accross
+    if (append) {
+      def moveFiles(src: Path, dst: Path): Hdfs[Unit] = for {
+        files <- Hdfs.files(src, "*.parquet")
+        time  =  System.currentTimeMillis
+        _     <- Hdfs.mkdirs(dst)
+        _     <- files.zipWithIndex.traverse {
+                   case (file, idx) => Hdfs.move(file, new Path(dst, f"part-$time-$idx%05d.parquet"))
+                 }
+      } yield ()
+
+      setup.flatMap(dst => Execution.fromHdfs(Hdfs.createTempDir()).bracket(
+        tmpDir => Execution.fromHdfs(Hdfs.delete(tmpDir, true))
+      )(
+        tmpDir => for {
+          counters <- write(tmpDir)
+          _        <- Execution.fromHdfs(moveFiles(tmpDir, dst))
+        } yield counters
+      ))
+    } else {
+      for {
+        dst      <- setup
+        counters <- write(dst)
+      } yield counters
+    }
   }
 
   override def write(pipe: TypedPipe[A], append: Boolean = true)
@@ -125,13 +163,13 @@ object HiveTable {
   def apply[A <: ThriftStruct : Manifest, B : Manifest : TupleSetter](
     database: String, table: String, partition: Partition[A, B], path: String
   ): HiveTable[A, (B,A)] =
-    new PartitionedHiveTable(database, table, partition, Some(path))
+    PartitionedHiveTable(database, table, partition, Some(path))
 
   /** Information need to address/describe a specific partitioned hive table.*/
   def apply[A <: ThriftStruct : Manifest, B : Manifest : TupleSetter](
     database: String, table: String, partition: Partition[A, B], pathOpt: Option[String] = None
   ): HiveTable[A, (B, A)] =
-    new PartitionedHiveTable(database, table, partition, pathOpt)
+    PartitionedHiveTable(database, table, partition, pathOpt)
 
   /** Information need to address/describe a specific unpartitioned hive table.*/
   def apply[A <: ThriftStruct : Manifest](
@@ -143,5 +181,5 @@ object HiveTable {
   def apply[A <: ThriftStruct : Manifest](
     database: String, table: String, pathOpt: Option[String] = None
   ): HiveTable[A, A] =
-    new UnpartitionedHiveTable(database, table, pathOpt)
+    UnpartitionedHiveTable(database, table, pathOpt)
 }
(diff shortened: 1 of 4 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,49 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
 ) extends HiveTable[A, A] {
 
   override def source =
-    new HiveParquetScroogeSource[A](database, table, externalPath)
+    HiveParquetScroogeSource[A](database, table, externalPath)
 
   override def sink(append: Boolean = true) =
-    new HiveParquetScroogeSource[A](database, table, externalPath)
+    if (append) throw new Exception("APPEND IS CURRENTLY NOT SUPPORTED")
```
SamRoberts: what is the relationship between this restriction and the functionality you have enabled in this commit?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,49 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
...
+      Hive.getPath(database, table)
+    )
+
+    //If we are appending the files are written to a temporary location and then copied accross
+    if (append) {
+      def moveFiles(src: Path, dst: Path): Hdfs[Unit] = for {
+        files  <- Hdfs.files(src, "*.parquet")
+        time   =  System.currentTimeMillis
+        exists <- Hdfs.exists(dst)
+        _      <- if (exists) Hdfs.mkdirs(dst) else Hdfs.value(true)
```
SamRoberts: did you mean `!exists`?

In any case, **IIRC** I believe `mkdirs` should return true if the directory already exists, so you should just be able to replace these two lines with a single `Hdfs.mkdirs` call. It is a great failing on behalf of our and hadoop's documentation that you have to dig so far into the code to find out mkdir's behavior here.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,49 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
...
+      } yield ()
+
+      for {
+        dst      <- setup
+        tmpDir   <- Execution.fromHdfs(Hdfs.createTempDir())
+        counters <- pipe
+                      .writeExecution(ParquetScroogeSource[A](tmpDir.toString))
+                      .getAndResetCounters
+                      .map(_._2)
+        _        <- Execution.fromHdfs(moveFiles(tmpDir, dst) >> Hdfs.delete(tmpDir, true))
```
SamRoberts: we should use `recoverWith` or some similar function to delete `tmpDir` no matter whether the previous steps succeed or fail

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,49 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
...
-      .getAndResetCounters
-      .map(_._2)
+    val setup = Execution.fromHive(
+      Hive.createParquetTable[A](database, table, List.empty, externalPath.map(new Path(_))) >>
+      Hive.getPath(database, table)
+    )
+
+    //If we are appending the files are written to a temporary location and then copied accross
+    if (append) {
+      def moveFiles(src: Path, dst: Path): Hdfs[Unit] = for {
```
SamRoberts: I would prefer not to use a nested method unless we really have to

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,49 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
...
+        time   =  System.currentTimeMillis
+        exists <- Hdfs.exists(dst)
+        _      <- if (exists) Hdfs.mkdirs(dst) else Hdfs.value(true)
+        _      <- files.zipWithIndex.traverse {
+                   case (file, idx) => Hdfs.move(file, new Path(dst, f"part-$time-$idx%05d.parquet"))
+                 }
+      } yield ()
+
+      for {
+        dst      <- setup
```
SamRoberts: why is `setup` necesary now where it wasn't before? is it because we are using `ParquetScroogeSource` now instead of `HiveParquetScroogeSource`?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
stephanh: Yes, I'll change that.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
stephanh: It never worked and this just makes it obvious. 

I added the ability to append if you use Execution. Once we deprecate the old API I want to remove this method.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
stephanh: Ok.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
stephanh: Not sure how else to structure the code without it becoming convoluted.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
SamRoberts: Cool

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
stephanh: Yes. We could just use `HiveParquetScroogeSource` here, but I thought it would be better to have it be as similar to the append case as possible so the chances of unexpected bevahiour are less.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
SamRoberts: what stops you from just defining `moveFiles` outside the `writeExecution` method? You're not even capturing any of the variables in the `writeExecution` method, unless I'm blind.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
SamRoberts: Oh, that's absolutely fine. I don't mind the change at all, I just wanted to make sure I understood the rationale.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
stephanh: `moveFiles` is rather specific to just this method. At this stage it doesn't make much sense standalone.
 
My preference would be for it to be part of the for comprehension but that was too convoluted.

SamRoberts: Ok, it's mostly just a matter of personal taste. I don't think there's anything wrong with moving blocks of functionality outside a method, but I also don't think there's anything wrong with using  nested method.

SamRoberts:
Once the `Hdfs.mkdirs` line is fixed, and the `tmpDir` deletion is protected inside the execution equivalent of a finally block, I am happy: :+1:

stephanh:
Getting `tmpDir` protected took quite a bit of effort.

@SamRoberts what do you think of this approach. Unfortunately, I ended up adding another nested method :(.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,51 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
...
+        files <- Hdfs.files(src, "*.parquet")
+        time  =  System.currentTimeMillis
+        _     <- Hdfs.mkdirs(dst)
+        _     <- files.zipWithIndex.traverse {
+                   case (file, idx) => Hdfs.move(file, new Path(dst, f"part-$time-$idx%05d.parquet"))
+                 }
+      } yield ()
+
+      setup.flatMap(dst => Execution.fromHdfs(Hdfs.createTempDir()).bracket(
+        tmpDir =>Execution.fromHdfs(Hdfs.delete(tmpDir, true))
```
SamRoberts: space between `=>` and `Execution`

stephanh: Fixed

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
```diff
@@ -100,18 +105,51 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
...
+      def moveFiles(src: Path, dst: Path): Hdfs[Unit] = for {
+        files <- Hdfs.files(src, "*.parquet")
+        time  =  System.currentTimeMillis
+        _     <- Hdfs.mkdirs(dst)
+        _     <- files.zipWithIndex.traverse {
+                   case (file, idx) => Hdfs.move(file, new Path(dst, f"part-$time-$idx%05d.parquet"))
+                 }
+      } yield ()
+
+      setup.flatMap(dst => Execution.fromHdfs(Hdfs.createTempDir()).bracket(
```
SamRoberts: I think this is good enough. Are you worried that it is not readable?

I think you could make it a little more readable by doing:
```
val withTmpDir = Execution.fromHdfs(Hdfs.createTempDir) bracket Execution.fromHdfs(Hdfs.delete(_, true))
setup.flatMap(dst => withTmpDir(tmpDir => for {
  counters <- write(tmpDir)
  _             <- Execution.fromHdfs(moveFiles(tmpDir, dst))
} yield counters ))

SamRoberts:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/hive/HiveTable.scala
SamRoberts: But, to be clear, I find it readable enough as it is

stephanh: I'll leave it as is. It's not the most readable but it is reasonable.

PR #292, opened 2015-01-08
Remove old API

SamRoberts:
This commit:
 - removes the old API
 - removes any associated examples
 - adds a TransformExecution example to replace the old Transform example
 - removes any associated code, tests, and resources
 - removes the exec namespace from the path of the new API
 - makes any associated scaladoc changes
 - refactors maestro's internals so it makes sense without the new API

This commit does not (we still need to):
 - change the README to reflect the removal of the old API


maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
```diff
@@ -17,34 +17,101 @@ package au.com.cba.omnia.maestro
 import com.twitter.scrooge.ThriftStruct
 
 package object api {
+  type MaestroJob      = au.com.cba.omnia.maestro.core.scalding.MaestroJob
+  type MacroSupport    = au.com.cba.omnia.maestro.macros.MacroSupport
+
+  type JobStatus       = au.com.cba.omnia.maestro.core.scalding.JobStatus
+  type JobFinished     = au.com.cba.omnia.maestro.core.scalding.JobFinished.type
+  type JobNotReady     = au.com.cba.omnia.maestro.core.scalding.JobNotReady.type
+  type JobNeverReady   = au.com.cba.omnia.maestro.core.scalding.JobNeverReady.type
```
stephanh: Not sure if we want/need to expose these types. I would expect our users to only write methods that have `JobStatus` or maybe `JobFailure` in their type signature.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
```diff
@@ -17,34 +17,101 @@ package au.com.cba.omnia.maestro
...
+  val  JobFinished     = au.com.cba.omnia.maestro.core.scalding.JobFinished
+  val  JobNotReady     = au.com.cba.omnia.maestro.core.scalding.JobNotReady
+  val  JobNeverReady   = au.com.cba.omnia.maestro.core.scalding.JobNeverReady
+  val  JobFailure      = au.com.cba.omnia.maestro.core.scalding.JobFailure
+
+  type HiveTable[A <: ThriftStruct, B] = au.com.cba.omnia.maestro.core.hive.HiveTable[A, B]
   type Partition[A, B] = au.com.cba.omnia.maestro.core.partition.Partition[A, B]
+  val  HiveTable       = au.com.cba.omnia.maestro.core.hive.HiveTable
+  val  Partition       = au.com.cba.omnia.maestro.core.partition.Partition
+  val  HivePartition   = au.com.cba.omnia.maestro.core.partition.HivePartition
```
stephanh: Should we merge Partition and HivePartition/get rid of one of them?

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
```diff
@@ -17,34 +17,101 @@ package au.com.cba.omnia.maestro
...
+  val  UploadInfo      = au.com.cba.omnia.maestro.core.task.UploadInfo
+
+  type LoadConfig[A]   = au.com.cba.omnia.maestro.core.task.LoadConfig[A]
+  type LoadInfo        = au.com.cba.omnia.maestro.core.task.LoadInfo
+  type LoadSuccess     = au.com.cba.omnia.maestro.core.task.LoadSuccess
+  type LoadFailure     = au.com.cba.omnia.maestro.core.task.LoadFailure
+  val  LoadConfig      = au.com.cba.omnia.maestro.core.task.LoadConfig
+  val  LoadSuccess     = au.com.cba.omnia.maestro.core.task.LoadSuccess
+  val  LoadFailure     = au.com.cba.omnia.maestro.core.task.LoadFailure
+  val  EmptyLoad       = au.com.cba.omnia.maestro.core.task.EmptyLoad
```
stephanh: Again, I'm wondering if we should expose all these. Some of them are needed for pattern matching which makes me wonder if we should have a `fold` on `LoadInfo`.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
```diff
@@ -17,34 +17,101 @@ package au.com.cba.omnia.maestro
...
+  val  SqoopImportConfig                                   = au.com.cba.omnia.maestro.core.task.SqoopImportConfig
+  val  SqoopExportConfig                                   = au.com.cba.omnia.maestro.core.task.SqoopExportConfig
+
+  val  ParlourInstances                       = au.com.cba.omnia.maestro.core.task.ParlourInstances
+  implicit val parlourImportDslMonoid         = ParlourInstances.parlourImportDslMonoid
+  implicit val parlourExportDslMonoid         = ParlourInstances.parlourExportDslMonoid
+  implicit val teradataParlourImportDslMonoid = ParlourInstances.teradataParlourImportDslMonoid
+  implicit val teradataParlourExportDslMonoid = ParlourInstances.teradataParlourExportDslMonoid
+
+  val  SqoopExecutionTest = au.com.cba.omnia.maestro.core.task.SqoopExecutionTest
```
stephanh: Test helpers usually go into `maestro-test`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
```diff
@@ -236,18 +251,76 @@ object SqoopEx extends Sqoop {
     } yield (importPath, count)
   }
 
+  def archive[C <: CompressionCodec : ClassManifest](
+    src: String, dest: String
+  ): Execution[Long] =
+    Execution.getConfigMode.flatMap { case (config, mode) =>
```
stephanh: Can we use `withSubConfig` from RichExecution here instead?

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: We talked about this when adding the original API, and this is what we went with, although neither of us had any strong arguments either way. Happy to take them out, or leave them in. I don't really see the advantage of taking them out though, unless you believe we are likely to remove some of these options?

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: They are actually different though: the partitioning patterns are different. It seems better to keep them separate for now.

If we want to force people to use the right partition for hive vs. non-hive, than one day we should probably create different types for `Partition` and `HivePartition`, and then having this split would be quite useful.

Your thoughts?

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: Why prefer a fold to pattern matching? pattern matching can be less convenient when the constructors are ugly (e.g. anything involving scalaz disjunction), but in general pattern matching is more powerful (hierarchical matches, guard clauses, partial functions, etc).

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: Ok, will move it

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
SamRoberts: I can change it, but isn't that more of a separate issue? I don't want to fix up all our execution coding issues in this PR just because my code moves the relevant files around.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
stephanh: They are different but in reality for HiveTable we just ignore the path component and otherwise we don't use the non Hive paths.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
stephanh: `fold` provides stronger case coverage checks. Also it allows us to hide the internals a bit more. In the case of a Fold you only need to know about `LoadInfo`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
stephanh: Sure, can you raise an issue for it please.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
```diff
@@ -17,34 +17,101 @@ package au.com.cba.omnia.maestro
...
+  type SqoopImportConfig[T <: ParlourImportOptions[T]]     = au.com.cba.omnia.maestro.core.task.SqoopImportConfig[T]
+  type SqoopExportConfig[T <: ParlourExportOptions[T]]     = au.com.cba.omnia.maestro.core.task.SqoopExportConfig[T]
+  val  SqoopImportConfig                                   = au.com.cba.omnia.maestro.core.task.SqoopImportConfig
+  val  SqoopExportConfig                                   = au.com.cba.omnia.maestro.core.task.SqoopExportConfig
+
+  val  ParlourInstances                       = au.com.cba.omnia.maestro.core.task.ParlourInstances
+  implicit val parlourImportDslMonoid         = ParlourInstances.parlourImportDslMonoid
+  implicit val parlourExportDslMonoid         = ParlourInstances.parlourExportDslMonoid
+  implicit val teradataParlourImportDslMonoid = ParlourInstances.teradataParlourImportDslMonoid
+  implicit val teradataParlourExportDslMonoid = ParlourInstances.teradataParlourExportDslMonoid
```
stephanh: Usually, we don't put implicits into the package object but instead put them into the correspondingly named camel case object i.e `Maestro`.

Importing the package should be completely safe and predictable. Importing the actual object should add the extra functionality.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: `fold` should not allow you to hide the internals more, though. The whole point of calling it a fold is to say that you are doing a fold on the structure of the class. Granted, there is no law upholding that, but if you keep changing the internals of your class while not changing the `fold`, you don't really have a fold anymore ... 

The scala compiler can complain about exhaustiveness checks too. We don't enable this? We should if we can get away with it.

I think it is desirable to have the user able to specify functions that only work on `LoadFailure` or `LoadSuccess`, etc. In fact, most executions that do a load should return `LoadSuccess`, not `LoadInfo`. Assuming the author of that execution hasn't cheated, this tells the user, just by looking at the type signature of the execution, that 1) the execution includes a load in it's success path, and 2) the execution fails if the load doesn't succeed. If the execution just returns `LoadInfo` then the user can't tell (2) without looking at the code.

stephanh:
Looks good but please hold off on merging it. It changes all the imports so we should check against the psgl load to see what impact it has so we can educate our users and ensure there is nothing unexpected.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: These are required to use the Dsl classes at all. The user shouldn't have to import `Maestro._` to be able to do that. Really, they should be in the companion object for their respective classes in parlour.

Maybe I will do that.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
SamRoberts: Sure, will do

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: Ok, that sounds good. But I think I will raise an issue for it rather than fix it now.

SamRoberts:
Sure, will hold off. But this diff is going to require manual changes almost every time it has to be rebased, simply because it touches lots of things. So if we can not wait too long, that would be nice. :)

stephanh:
I know. I'll test it against psgl now. Can we also change the version number to say `1.99`, then we do some of the other fixes and bump the version to `2.0`?

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: Issue raised

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
```diff
@@ -17,34 +17,101 @@ package au.com.cba.omnia.maestro
...
+
+  val  SqoopExecutionTest = au.com.cba.omnia.maestro.core.task.SqoopExecutionTest
 }
```
stephanh: I think we should add the `ThriftStruct` type to package. I'm not aware of any load that imports anything from Scrooge apart from ThriftStruct.

SamRoberts: Good idea. But what about defining a new `thirdparty` or `external` package object for this type, and `Hdfs` and `Hive`?

stephanh: Not convinced :smile:.

SamRoberts: Thinking about it some more, I agree with your disagreement.

If we have no such package object, and people want to import _ from a third party lib, they need to mask the import of that type from this package.

If they import the `thirdparty` package object, and they also import _ from a third party lib, they still need to mask the import of that type from the `thirdparty` package. The only advantage of this scheme is that there are less types to wade through in the `thirdparty` package, but that probably won't make any real difference in practice.

In order to avoid the user having to mask imports of specific types, we would need to have a separate package object for each dependency, which seems a bit over the top.

TL:DR will add `ThiftStruct` to the api package object.

stephanh:
:+1:

I tested the impact these changes have on PSGL and it's fine so I'm happy for this get merged when you are done.

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: Done

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
SamRoberts: Raised

maestro-api/src/main/scala/au/com/cba/omnia/maestro/api/package.scala
SamRoberts: Done

SamRoberts:
yay! broke the 2200 net lines removed barrier!

PR #296, opened 2015-01-12
Minor API cleanup

stephanh:
* Removes `write` and `sink` from HiveTable since they are broken and unused with the new API.
* Refactors SqoopExecution to use `withSubConfig`.


SamRoberts:
:+1:

PR #297, opened 2015-01-13
Multiple sqoop queries fix

### LINKED ISSUE #286: Should set the class name when running sqoop
Unfortunately, when multiple sqoop jobs are run in the same JVM (such as when running tests) it caches some state from the first test. This can be mitigated by setting the class name to different names for each test.

stephanh:
When importing data using Sqoop using a query Sqoop creates a java class and jar with a generic name. If there are 2 sqoop imports in the same job they will have the same name and clash causing the second import to fail.

This sets a unique class name for each sqoop job.

This closes #286.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
@@ -14,6 +14,8 @@
 
 package au.com.cba.omnia.maestro.core.task
 
+import scala.util.Random
+
 import java.io.File
 
 import org.apache.commons.lang.StringUtils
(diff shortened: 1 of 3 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
```diff
@@ -234,12 +236,14 @@ object SqoopEx {
...
-    val withDestDir  = config.options.targetDir(importPath)
-    val withMRHome   = setCustomMRHome(withDestDir)
-    val sqoopOptions = withMRHome.toSqoopOptions
+    val importPath    = config.hdfsLandingPath + File.separator + config.timePath
+    val archivePath   = config.hdfsArchivePath + File.separator + config.timePath
+    val logger        = Logger.getLogger("Sqoop")
+    val withDestDir   = config.options.targetDir(importPath)
+    val withMRHome    = setCustomMRHome(withDestDir)
+    val withClassName =
+      withMRHome.getClassName.fold(withMRHome.className(s"SqoopImport_${Random.nextInt(Int.MaxValue)}"))(_ => withMRHome)
```
SamRoberts: Minor suggestion: if we have to look at multiple job numbers simultaneously in logs, it might be worth zero padding the number so it always takes up the same number of characters. But up to you.

SamRoberts:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
stephanh: Done.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/SqoopExecution.scala
```diff
@@ -275,8 +279,9 @@ object SqoopEx {
     val withDelete =
       if (config.deleteFromTable) trySetDeleteQuery(config.options)
       else config.options
-    val withMRHome = setCustomMRHome(withDelete)
-    ParlourExecution.sqoopExport(withMRHome)
+    val withMRHome    = setCustomMRHome(withDelete)
+    val withClassName = withMRHome.getClassName.fold(withMRHome.className(s"SqoopExport_${Random.nextInt(Int.MaxValue)}"))(_ => withMRHome)
```
rowandavies: Minor fussiness: line goes beyond 120 columns, and inconsistent with the break in line 245.

stephanh: You are right. I tried making one of them better and forgot about the other. I find that there is not really a nice way to format curried functions.

Issue #300, opened 2015-01-14
Scalding counter fails often


stephanh:
Not sure why but on some map tasks for some loads trying to access the `FlowDef` on the individual mapppers to increase the filter counter fails. Scalding stores the `FlowDef`s in a HashMap and uses the job id to look them up. Below is the execption:
```
2015-01-12 13:34:50,292 ERROR cascading.flow.stream.TrapHandler: caught Throwable, no trap available, rethrowing
cascading.pipe.OperatorException: [com.twitter.scalding.M...][com.twitter.scalding.RichPipe.eachTo(RichPipe.scala:478)] operator Each failed executing operation
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:107)
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:39)
	at cascading.flow.stream.FunctionEachStage$1.collect(FunctionEachStage.java:80)
	at cascading.tuple.TupleEntryCollector.safeCollect(TupleEntryCollector.java:145)
	at cascading.tuple.TupleEntryCollector.add(TupleEntryCollector.java:133)
	at au.com.cba.omnia.maestro.core.task.ExtractTime.operate(LoadHelper.scala:265)
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:99)
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:39)
	at cascading.flow.stream.SourceStage.map(SourceStage.java:102)
	at cascading.flow.stream.SourceStage.run(SourceStage.java:58)
	at cascading.flow.hadoop.FlowMapper.run(FlowMapper.java:130)
	at org.apache.hadoop.mapred.MapTask.runOldMapper(MapTask.java:417)
	at org.apache.hadoop.mapred.MapTask.run(MapTask.java:332)
	at org.apache.hadoop.mapred.Child$4.run(Child.java:268)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:396)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1438)
	at org.apache.hadoop.mapred.Child.main(Child.java:262)
Caused by: java.lang.RuntimeException: Error in job deployment, the FlowProcess for unique id UniqueID(1490058012982771712) isn't available
	at scala.sys.package$.error(package.scala:27)
[...]

PR #301, opened 2015-01-15
Uses cascading-counter for filtering instead of scalding.

stephanh:
Work around #300 until we figure out what the problem is and can fix it.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
@@ -179,19 +179,30 @@ object LoadEx {
 
   def parseRows[A <: ThriftStruct : Decode : Tag : Manifest](
     conf: LoadConfig[A], filterCounter: Stat, in: TypedPipe[RawRow]
-  ): TypedPipe[String \/ A] =
-    in
-      .map(row => conf.splitter.run(row.line) ++ row.extraFields)
-      .flatMap(conf.filter.run(_) match {
+  ): TypedPipe[String \/ A] = {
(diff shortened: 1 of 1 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -179,19 +179,30 @@ object LoadEx {
...
+      )
+    }
+
+    pipe
+      .map { case (row, fp) => (conf.splitter.run(row.line) ++ row.extraFields, fp) }
+      .flatMap { case (parts, fp) => conf.filter.run(parts) match {
         case Some(r) => Some(r)
         case None    => {
-          filterCounter.inc
+          fp.increment("maestro", "tuples_filtered", 1L)
```
SamRoberts: what is going to happen if there are multiple load tasks in the one execution?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -179,19 +179,30 @@ object LoadEx {
...
+          //filterCounter.inc
```
SamRoberts: delete commented out code?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
stephanh: It should be fine. The job counts only a apply to a particular scalding run.

The constraints around this in Scalding are not well typed but instead require you to do the right thing at the right time or it will fail.

SamRoberts:
assuming you do whatever is appropriate with your commented out line: :+1:

PR #320, opened 2015-02-12
The load return pipe provides a smaller number of larger splits

SamRoberts:
We also switch to using a compact encoding for the temp sequence files.

The load pipe split size is configurable.

NOTE: I have not yet tested this on the cluster against a real load, but will do so before merging to master.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
@@ -16,10 +16,15 @@ package au.com.cba.omnia.maestro.core.task
 
 import java.nio.ByteBuffer
 import java.security.{MessageDigest, SecureRandom}
+import java.util.UUID
 
+import scala.reflect.runtime.currentMirror
+import scala.util.control.NonFatal
 import scala.util.hashing.MurmurHash3
 
 import scalaz.{Tag => _, _}, Scalaz._
+
+import org.apache.hadoop.io.{BytesWritable, NullWritable}
 
 import cascading.flow.FlowProcess
 import cascading.operation.{BaseOperation, Function, FunctionCall, OperationCall}
@@ -30,12 +35,14 @@ import com.twitter.scalding.{Execution, ExecutionCounters, MultipleTextLineFiles
 import com.twitter.scalding.typed.TypedPipeFactory
 import com.twitter.scalding.Dsl._ // Pipe.eachTo
 
-import com.twitter.scrooge.ThriftStruct
+import com.twitter.bijection.scrooge.CompactScalaCodec
+
+import com.twitter.scrooge.{ThriftStruct, ThriftStructCodec}
 
 import au.com.cba.omnia.maestro.core.codec._
 import au.com.cba.omnia.maestro.core.clean.Clean
 import au.com.cba.omnia.maestro.core.filter.RowFilter
-import au.com.cba.omnia.maestro.core.scalding.{Errors, StatKeys}
+import au.com.cba.omnia.maestro.core.scalding._
 import au.com.cba.omnia.maestro.core.split.Splitter
 import au.com.cba.omnia.maestro.core.time.TimeSource
 import au.com.cba.omnia.maestro.core.validate.Validator
@@ -120,6 +127,9 @@ object LoadInfo {
   *                        Defaults to no validation.
   * @param errorThreshold: The fraction of rows that can fail during the load step
   *                        before the overall job should fail. Defaults to 0.05 or 5%.
+  * @param splitSize:      The size of each split in the job that processes the
+  *                        pipe returned by load. Defaults to 9 blocks, with the
+  *                        hope that this will result in 2-block parquet files.
   */
 case class LoadConfig[A](
   errors: String,
@@ -130,7 +140,8 @@ case class LoadConfig[A](
   clean: Clean            = Clean.default,
   none: String            = "",
   validator: Validator[A] = Validator.all[A](),
-  errorThreshold: Double  = 0.05
+  errorThreshold: Double  = 0.05,
+  splitSize: Long         = 9L * 128 * 1024 * 1024
 )
 
 /** Executions for load tasks */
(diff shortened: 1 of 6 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -166,15 +178,23 @@ object LoadEx {
       if (config.generateKey) pipeWithDateAndKey(sources, config.timeSource)
       else                    pipeWithDate(sources, config.timeSource)
 
-    Execution.withId { id =>
+    Execution.withId { id => Paths.tempDir.flatMap { tmpDir => {
```
vineethvarghese: Do you think getting the temp directory of a job might be something useful or common enough to add to the RichExecution?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -130,7 +140,8 @@ case class LoadConfig[A](
   clean: Clean            = Clean.default,
   none: String            = "",
   validator: Validator[A] = Validator.all[A](),
-  errorThreshold: Double  = 0.05
+  errorThreshold: Double  = 0.05,
+  splitSize: Long         = 9L * 128 * 1024 * 1024
```
vineethvarghese: May not be important now, but would be good we have performance numbers around this so that people can make well informed decisions when they want to pick a value.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
SamRoberts: I'm not sure if it would be a good thing to encourage our users to put files into the temp dir, unless perhaps we were to supply some guarantees around removing those files when the job finishes (which would be nice). We're already running into problems with the temp dir filling up as it is.

Even if we did expose it to users, I'm not sure I would want to cram every single extra execution we provide into `RichExecution`. I think that makes a lot of sense for the `fromBlah` functions. But not functions which have a clearly identifiable domain that they could belong too instead.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
SamRoberts: Sure. I think this is going to have to be something we learn the hard way.

vineethvarghese: We haven't done any performance test on Maestro as far as I know. It would be good to have some automated performance in the future.

vineethvarghese:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -166,15 +178,23 @@ object LoadEx {
...
+      val serializer  = CompactScalaCodec[A](getCodec[A])
+      val seqFileIn   = thriftInput.map(t => (NullWritable.get, new BytesWritable(serializer(t))))
+      val seqFileOut  = TypedPipe.from[(NullWritable, BytesWritable)](seqSource)
+                          .map { case (_, bytes) => serializer.invert(bytes.getBytes).get }
+
+      seqFileIn.writeExecution(seqSource).getAndResetCounters.map { case (_, counters) =>
+        (seqFileOut, LoadInfo.fromCounters(counters, config.errorThreshold))
       }
-    }
+    }}}
```
vineethvarghese: This was confusing for a second.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/LoadViewExecutionSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import au.com.cba.omnia.maestro.core.clean.Clean
+import au.com.cba.omnia.maestro.core.filter.RowFilter
+import au.com.cba.omnia.maestro.core.hive.HiveTable
+import au.com.cba.omnia.maestro.core.partition.Partition
+import au.com.cba.omnia.maestro.core.scalding.ExecutionOps._
+import au.com.cba.omnia.maestro.core.split.Splitter
+import au.com.cba.omnia.maestro.core.time.TimeSource
+import au.com.cba.omnia.maestro.core.validate.Validator
+
```
vineethvarghese: Why do we have this space?

SamRoberts: Yeah. The above imports are part of the maestro core production code base. The below import is just the thrift file for these tests. They seemed conceptually distinct to me. But I'm happy either way.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
SamRoberts: Agreed.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
SamRoberts: Hah! The opening line for those braces is pretty dense, with the `withId` and `flatMap` call on the same line. Do you think I should change it?

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/LoadViewExecutionSpec.scala
vineethvarghese: If that is the pattern we were following(I can't remember), then it should be fine.

vineethvarghese:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
vineethvarghese: It did take me a second to see that but I guess it is better than 4 levels of indentation with little value. So, lets leave that.

PR #323, opened 2015-02-20
Generate key path fix

rowandavies:
This fix just always expands the `sources` passed to `load` to the full paths for all files.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
@@ -47,6 +47,9 @@ import au.com.cba.omnia.maestro.core.scalding._
 import au.com.cba.omnia.maestro.core.split.Splitter
 import au.com.cba.omnia.maestro.core.time.TimeSource
 import au.com.cba.omnia.maestro.core.validate.Validator
+import au.com.cba.omnia.maestro.core.scalding.ExecutionOps._
+
+import au.com.cba.omnia.permafrost.hdfs.Hdfs
 
 /** Information about a Load */
 sealed trait LoadInfo {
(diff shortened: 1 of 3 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -161,8 +165,25 @@ trait LoadExecution {
...
+
+    for {
+      srcsResolved <- Execution.fromHdfs(
+                        for {
+                          fs       <- Hdfs.filesystem
+                          paths    =  sources.map(Hdfs.path(_))
+                          resPaths <- paths.map(resolvePath).sequence
+                        } yield resPaths.flatten
+                      )
+      (pipe, info) <- LoadEx.execution[A](config, srcsResolved.map(_.toString))
```
SamRoberts: As far as I can see, this logic should be inside the execution function.

SamRoberts:
happy to approve, with the proviso that we continue to work on this, fix the test, refactor the code, etc.

:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -161,8 +165,25 @@ trait LoadExecution {
     */
   def load[A <: ThriftStruct : Decode : Tag : Manifest](
     config: LoadConfig[A], sources: List[String]
-  ): Execution[(TypedPipe[A], LoadInfo)] =
-    LoadEx.execution[A](config, sources)
+  ): Execution[(TypedPipe[A], LoadInfo)] = {
+    val resolvePath = (srcPath: Path) =>
```
stephanh: Minor nitpick but I think just using a `def` for situations like this would be nicer.
`` def resolvePath(src: Path) = ```

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/LoadExecutionSpec.scala
```diff
@@ -87,4 +88,13 @@ Load execution properties
       executesSuccessfully(exec)._2 must_== LoadSuccess(5, 3, 3, 0)
     }
   }
+
+  // This checks for lack of exceptions, but not success - a new thrift type is required.
```
stephanh: Why is a new thrift struct required?

rowandavies: Because an extra field for the generated key is added.

PR #327, opened 2015-02-24
Sam upload misc fields

SamRoberts:
Adds the ability to surround wildcards in file patterns with braces. E.g. `{table}_date_{yyyyMMdd}_seq{??}.DAT`. The values in these "captured" wildcards are appended to the list of directories that upload stores files in, thereby ensuring that file with different captured values are uploaded to different folders.

I had hoped to have time to clean up my old code in this file, but unfortunately I ran out of time before making any non-superficial changes. Still, at least it's no worse than it was before.


SamRoberts:
Added (in a bit of a hurry) a more substantial "clean up" commit: breaking up the logic in InputParser into three main separate bins. Unfortunately these bins form a stack of traits, so not 100% sure this is an improvement over having everything together.

Anyway, it might be easier to review the first four commits together, and then review the last commit. As the last commit moves all the code around, so you lose sight of what specific changes were made to the logic. Could drop that last commit and move it into a separate PR if you think it is easier.

stephanh:
Why do all the files need to go into separate files?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/FileNameMatchers.scala
```diff
@@ -0,0 +1,236 @@
...
+/** The file does not match the pattern */
+case object NoMatch extends MatchResult
+
+/**
+  * The file matches the pattern.
+  *
+  * Also contains the directories for the file
+  * corresponding to the fields in the pattern.
+  */
+case class Match(dirs: List[String]) extends MatchResult
```
stephanh: I know this is old code but confused about `dirs`. Is it something like `List("usr", "local", "bin")` where the different directories in a path are broken up?

SamRoberts:
Instead of adding this new feature to the file pattern, we are going to fix maestro so it can upload multiple files into the one directory

PR #330, opened 2015-02-26
upload can handle multiple sequence files writing to the same directory

SamRoberts:


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -166,13 +166,13 @@ object UploadEx {
...
+    copied <- Execution.fromHdfs(Push.push(files.dataFiles, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath))
 
     _      <- Execution.from {
+      copied.foreach { case Copied(files, dest) =>
+        logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ")
+      }
       logger.info(s"Upload ended from ${conf.localIngestPath}")
     }
-  } yield UploadInfo(copied)
+  } yield { UploadInfo(copied.map(_.dest.toString)) }
```
stephanh: The `{}` are superfluous.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -166,13 +166,13 @@ object UploadEx {
...
     ))
 
-    copied <- Execution.fromHdfs(files.dataFiles.traverse(data => for {
-      record <- Push.push(data, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath)
-      _      =  logger.info(s"copied ${record.source.getName} to ${record.dest}")
-    } yield record.dest.toString ))
+    copied <- Execution.fromHdfs(Push.push(files.dataFiles, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath))
 
     _      <- Execution.from {
+      copied.foreach { case Copied(files, dest) =>
```
stephanh: Can we indent this  under the `Execution`.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/PushSpec.scala
```diff
@@ -26,7 +27,7 @@ import au.com.cba.omnia.omnitool.{Error, Ok}
 
 import au.com.cba.omnia.permafrost.hdfs.Hdfs
 
-class PushSpec extends Specification { def is = s2"""
+class PushSpec extends Specification with ThrownExpectations { def is = s2"""
```
stephanh: Should you use theremomter spec here or at least `HadoopSupport` to provide a good conf file?

stephanh:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: will fix

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: will fix

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/PushSpec.scala
SamRoberts: These tests wouldn't make good use of HadoopSupport anyway. I find these tests a maintenance nightmare. I think, if I were to do them again, I would replace them all with a couple of well chosen scalacheck tests. (Push has some properties it should maintain around directory structures, preserving and archiving files, etc.) And yeah, probably using HadoopSupport to run the Hdfs actions would be part of that.

PR #332, opened 2015-03-16
Adds a Validator that never fails.

### LINKED ISSUE #211: Add a validator that never fails
Currently users write `Validator.all[T]()` when they want to not do any validation, meaning run all of these 0 validators on the input. This is slightly counter-intuitive. Would prefer to have a single Validator value that never fails.

stephanh:
This closes #211.


SamRoberts:
:+1:

PR #336, opened 2015-03-17
CDH5 upgrade

stephanh:
CDH5 upgrade and force view execution to create the hive database.
This prevents the concurrent attempt to create the database within
HiveTap from failing and crashing the job.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/ViewExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/ViewExecution.scala
@@ -19,10 +19,12 @@ import com.twitter.scalding.{Execution, TupleSetter, TypedPipe}
 import com.twitter.scrooge.ThriftStruct
 
 import au.com.cba.omnia.ebenezer.scrooge.PartitionParquetScroogeSource
+import au.com.cba.omnia.ebenezer.scrooge.hive.Hive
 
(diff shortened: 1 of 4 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/ViewExecution.scala
```diff
@@ -61,7 +63,8 @@ trait ViewExecution {
     */
   def viewHive[A <: ThriftStruct : Manifest, ST](
     table: HiveTable[A, ST], pipe: TypedPipe[A], append: Boolean = true
-  ): Execution[Long] =
-    table.writeExecution(pipe, append)
-      .map(_.get(StatKeys.tuplesWritten).getOrElse(0))
+  ): Execution[Long] = for {
+    _ <- Execution.fromHive(Hive.createDatabase(table.database))
```
SamRoberts: I think we could do with a comment in the source code to explain why this is necesary

SamRoberts:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/ViewExecution.scala
stephanh: Good idea :smile:.

stephanh: Done.

Issue #321, opened 2015-02-17
Setting the property loadWithKey to true within LoadConfig caused the Job to fail


MadhuMi:
When i  set the property loadWithKey to true within LoadConfig to test jobs that require the automatically generated row_id field as the last value , the job failed with following exception :-

cascading.pipe.OperatorException: [com.twitter.scalding.M...][com.twitter.scalding.RichPipe.eachTo(RichPipe.scala:478)] operator Each failed executing operation
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:107)
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:39)
	at cascading.flow.stream.SourceStage.map(SourceStage.java:102)
	at cascading.flow.stream.SourceStage.run(SourceStage.java:58)
	at cascading.flow.hadoop.FlowMapper.run(FlowMapper.java:130)
	at org.apache.hadoop.mapred.MapTask.runOldMapper(MapTask.java:417)
	at org.apache.hadoop.mapred.MapTask.run(MapTask.java:332)
	at org.apache.hadoop.mapred.Child$4.run(Child.java:268)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:396)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1438)
	at org.apache.hadoop.mapred.Child.main(Child.java:262)
Caused by: java.util.NoSuchElementException: key not found: hdfs://nameservice1/dev/source/HLS/AMORTISATIONTRAN_SZGLAMR/2015/01/30/I5_OMN_HLS_SZGLAMR_20150130.DAT
	at scala.collection.MapLike$class.default(MapLike.scala:228)
	at scala.collection.AbstractMap.default(Map.scala:58)
	at scala.collection.MapLike$class.apply(MapLike.scala:141)
	at scala.collection.AbstractMap.apply(Map.scala:58)
	at au.com.cba.omnia.maestro.core.task.GenerateKey.uid(LoadExecution.scala:263)
	at au.com.cba.omnia.maestro.core.task.GenerateKey.operate(LoadExecution.scala:297)
	at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:99)
	... 11 more

SamRoberts:
[...]

Issue #147, opened 2014-08-15
Load ignores too much input


stephanh:
The Load command silently drops columns that exceeded the number of fields in the thrift struct.

The Tag.tag command will just force each input row to be the right length.

stephanh:
I can fix this problem by either changing `case class Tag[A](run: List[String] => List[(String, Field[A, _])])` to something like `case class Tag[A](run: List[String] => String \/ List[(String, Field[A, _])])` and ensuring that the run method fails if the list is too short or long. This will have an impact on backwards compatibility. Probably not too much since I don't think anyone use it directly.

Or I can just add a check in `LoadEx.parseRows` to discover this error and handle it there.

Thoughts anyone?

SamRoberts:
I was imagining the first method. Where reasonable, I would prefer to report errors at the point where they arise, not check for errors first and then rely on that check for the correctness of later steps. But I was assuming that users weren't using tag directly.

(Which is a good point? If some of these "internal" load types are in our api, maybe we should take them out?)

vineethvarghese:
Changing the run on Tag seems the right way if possible.

rowandavies:
@stephanh Maybe I've missed something, but I thought that in the existing `LoadEx.parseRows`, within the `Decode.decode` will already catch the situation of extra fields and hit the `case TooMuchInput =>`. 

But, I'd agree that the first solution is better.  Defining the tags for `StringPair` and `StringTriple` as infinite streams doesn't feel so natural.

SamRoberts:
@rowandavies the current implementation of tag, created by our macros, zips the list of columns with the list of expected `Field`s, thereby truncating any extra columns before they are ever passed to decode.

rowandavies:
Oh.  So then the maestro tests using `StringPairTag` aren't representative of production code.   Having run the tests, and seen the rows with extra columns get tagged as errors, I was mislead into thinking that was what would happen during actual loads.  So definitely something needs to be fixed here.

I don't object to the zip with an infinite stream, but I don't think this should be built into `StringPair` and `StringTriple`.  We could instead to add the infinite stream at the end when we do the zip via `++` or similar.  Or, we could make the errors explicit from Tag.tag.

stephanh:
Agreed. I want to deal with this properly and want StringPair and StringTiple to be representative of what they should be like.

stephanh:
Since no one spoke up about concerns with backwards compatibility and I don't think it will have a big impact I'm going to change the tag implementation.

[...]

PR #338, opened 2015-03-20
Much simpler fix for the generateKey bug #321 that just passes the seed in an...

rowandavies:
...d calculates the hash in `GenerateKey.prepare`.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
@@ -47,9 +47,6 @@ import au.com.cba.omnia.maestro.core.scalding._
 import au.com.cba.omnia.maestro.core.split.Splitter
 import au.com.cba.omnia.maestro.core.time.TimeSource
 import au.com.cba.omnia.maestro.core.validate.Validator
-import au.com.cba.omnia.maestro.core.scalding.ExecutionOps._
-
-import au.com.cba.omnia.permafrost.hdfs.Hdfs
 
 /** Information about a Load */
 sealed trait LoadInfo {
@@ -166,7 +163,7 @@ trait LoadExecution {
   def load[A <: ThriftStruct : Decode : Tag : Manifest](
     config: LoadConfig[A], sources: List[String]
   ): Execution[(TypedPipe[A], LoadInfo)] = {
-      LoadEx.execution[A](config, sources)
+    LoadEx.execution[A](config, sources)
   }
 }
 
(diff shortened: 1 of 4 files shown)


rowandavies:
This fixes #321 in a more robust way that's also much simpler.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -268,17 +251,11 @@ object LoadEx {
   def pipeWithDateAndKey(sources: List[String], timeSource: TimeSource): TypedPipe[RawRow] = {
     val rnd    = new SecureRandom()
```
vineethvarghese: Do we use this somewhere?

rowandavies: It's only used in the next line to generate the seed.

vineethvarghese:
:+1:

rowandavies:
@vineethvarghese I've updated the PR to extend the test to check that the generated keys are unique.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -336,24 +313,25 @@ class GenerateKey(timeSource: TimeSource, hashes: Map[String, String])
     */
   override def prepare(
```
stephanh: Prepare is called once per map task so in order for this to work as expected we need to ensure that there is no combining of blocks from multiple files into 1 task.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/StringTripleSupport.scala
```diff
@@ -0,0 +1,40 @@
...
+    second <- Decode.of[String]
+    third  <- Decode.of[String]
+  } yield StringTriple(first, second, third)
+
+  implicit val StringTripleTag: Tag[StringTriple] = {
+    val fields =
+      Field("FIRST",  (p:StringTriple) => p.first) +:
+      Field("SECOND", (p:StringTriple) => p.second) +:
+      Field("THIRD",  (p:StringTriple) => p.third) +:
+      Stream.continually[Field[StringTriple,String]](
```
stephanh: What is the purpose of this code?

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/StringTripleSupport.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  implicit val StringTripleDecode: Decode[StringTriple] = for {
+    first  <- Decode.of[String]
+    second <- Decode.of[String]
+    third  <- Decode.of[String]
+  } yield StringTriple(first, second, third)
+
+  implicit val StringTripleTag: Tag[StringTriple] = {
+    val fields =
+      Field("FIRST",  (p:StringTriple) => p.first) +:
```
stephanh: Missing space should be: `p: StringTriple`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
rowandavies: True.  I thought this was ensured because the original code only called `getProperty(CASCADING_SOURCE_PATH)` in `prepare` - so if there's an issue, it is an old issue.  Good to fix it if so.  @SamRoberts said he thought in this context each mapper would only involve a single file, but it's not clear to me why (Sam?), and I'm finding the cascading documentation and source code opaque on this point - any pointers?

As an alternative, it's trivial to move the call to `getProperty(CASCADING_SOURCE_PATH)` to `operate` if we think that won't have performance consequences.  The efficiency of the path hashing won't dominate, since we're hashing the entire row anyway.  But, if we wanted to be sure we could have memoisation of path hashes local to `prepare` and `operate`.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/StringTripleSupport.scala
rowandavies: It serves the same purpose as the similar code in `StringPairSupport.StringPairTag`, which seems to be to allow error rows to have extra fields without an exception being thrown - instead the error is caught elsewhere.  For this particular test it's not needed because there are no error rows.  But, having StringPair and StringTriple be inconsistent could cause confusion at a later point.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/StringTripleSupport.scala
rowandavies: Oops.  Fixed in commit `Style fixes` 2c05a36

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
SamRoberts: @rowandavies we talked about this on maestro chat. I raised the same issue that @stephanh raised (although mine was only a guess as to how often prepare was called), and you noted that the existing code already suffered from the same issue. I checked that, then agreed with you, and noted that we would never see the problem with maestro as it currently stands because the MR job that this code runs in does not aggregate blocks.

I think, given it is an existing issue, it would be reasonable to just add some comments in the code, log a maestro issue about it, and call it a day. If we are to look at it, I would like to see some tests that use aggregated map tasks so that we know for sure what the cascading source path property gives you in that case.

stephanh: Ok.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/StringTripleSupport.scala
stephanh: I rather have it inconsistent with `StringPair` and be correct/accurate.
Maybe we need to do a tidy up in StringPair to provide a custom tag implementation for the test.

SamRoberts: Hold on, not so fast!

The code created by our tag macro silently truncates columns to fit the fields expected by tag, rather then throwing an error (and by truncating, stops decode from throwing an error later too). This is an existing bug in maestro, which IIRC we already have an issue for.

This custom implementation is a slight tweak on the existing code produced by the macro, extending it with an infinite list of fields so that it won't silently truncate columns but will instead keep them around (meaning the decoder throws a meaningful error later). Although looking at it now, I can't remember why I didn't just add code to check the number of columns and throw an exception if it is wrong ...

Anyway, I think fixing these custom implementation is a decent idea, but it's far more important to fix the code generated by our macros, which is both far worse and actually going into production.

stephanh: @SamRoberts I agree with you that we need to fix that bug. However, considering that there is always a temptation to see something that works and just copy it. I don't want things like this to spread and then when we fix the original issue they are still around.

SamRoberts: Hrmm, thinking about it some more, I'm not convinced that it's better to throw an exception from the tag method. In the rest of LoadExecution, we have gone to some pains to identify those parts of the pipeline which can throw an exception and use flatMap and friends to convert them to disjunctions and pipe the errors to the errors file. The tag part of load assumes that tagging does not fail: that's wrong, but I don't think you want to change it in this PR. What real benefit is there in throwing that errors infrastructure away and throwing an exception in the tag method? The "zip on infinite stream" code above ensures that you get a meaningful error in the decode step.

(This was probably why I did it this way originally)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
rowandavies: Okay.  I've created issue #341

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/StringTripleSupport.scala
rowandavies: For the case of `StringTripleTag`, I guess I'm agreeing with @stephanh and I've removed the infinitely stuttering stream, since the current test that uses `StringTriple` never has extra fields anyway, and addressing #147 seems likely to make the point moot.

PR #340, opened 2015-03-23
Better stack trace for `RichExecution.from...` and adds `RichExecution.fromResult`

### LINKED ISSUE #326: Execution.from methods don't have enough information when failing
The Execution.from methods only include the exception of  the future that was supposed to be run but no information about the context in which the future is being run. So you don't actually know where in the code the failure occurred.

stephanh:
Changes the internals of `RichExecution` `fromHive`, `fromHdfs` and `fromEither` to provide a more useful stack trace. The stack trace now shows the location that `fromHdfs` or `fromHive` where called. This closes #326.

Also adds `fromResult`.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
@@ -95,6 +95,23 @@ case class RichExecutionObject(exec: Execution.type) {
     }
   }
 
+  /** Changes from an action that produces a Result to an Execution. */
+  def fromResult[T](result: => Result[T]): Execution[T] = {
+    /* Need to get the stack trace outside of the execution so we get a stack trace that is from the
+     * main thread which includes the stack that called this function.
+     */
+    val stacktrace = Thread.currentThread.getStackTrace.tail
+
+    Execution.fromFuture[T](_ => resultToFuture(result, "Operation failed", stacktrace))
+  }
+
+  /** Changes from an action that produces a scalaz Disjunction to an Execution. */
+  def fromEither[T](disjunction: => String \/ T): Execution[T] =
+    Execution.fromFuture(_ => disjunction.fold(
+      msg => Future.failed(new Exception(msg)),
+      x   => Future.successful(x)
+    ))
+
   /**
     * Helper function to convert a result to a future.
     *
@@ -108,22 +125,15 @@ case class RichExecutionObject(exec: Execution.type) {
       Future.successful,
       error => {
         val (msg, cause) = error match {
-          case This(msg)     => (s"$prefix: $msg", null)
-          case That(ex)      => (s"$prefix."     , ex)
-          case Both(msg, ex) => (s"$prefix: $msg", ex)
+          case This(msg)     => (s"$prefix: $msg"            , null)
+          case That(ex)      => (s"$prefix: ${ex.getMessage}", ex)
+          case Both(msg, ex) => (s"$prefix: $msg"            , ex)
         }
         val exception = new Exception(msg, cause)
         exception.setStackTrace(stacktrace)
         Future.failed(exception)
       }
     )
-
-  /** Changes from an action that produces a scalaz Disjunction to an Execution. */
-  def fromEither[T](disjunction: => String \/ T): Execution[T] =
-    Execution.fromFuture(_ => disjunction.fold(
-      msg => Future.failed(new Exception(msg)),
-      x   => Future.successful(x)
-    ))
 }
 
 object ExecutionOps extends ExecutionOps
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
@@ -107,10 +107,7 @@ case class RichExecutionObject(exec: Execution.type) {
 
   /** Changes from an action that produces a scalaz Disjunction to an Execution. */
   def fromEither[T](disjunction: => String \/ T): Execution[T] =
-    Execution.fromFuture(_ => disjunction.fold(
-      msg => Future.failed(new Exception(msg)),
-      x   => Future.successful(x)
-    ))
+    fromResult(disjunction.fold(Result.fail, Result.ok))
 
   /**
     * Helper function to convert a result to a future.
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
@@ -17,10 +17,13 @@ package au.com.cba.omnia.maestro.core.scalding
 import scala.concurrent.Future
 
 import scalaz._, Scalaz._
+import scalaz.\&/.{This, That, Both}
 
 import com.twitter.scalding.{Config, Execution}
 
 import org.apache.hadoop.hive.conf.HiveConf
+
+import au.com.cba.omnia.omnitool.Result
 
 import au.com.cba.omnia.permafrost.hdfs.Hdfs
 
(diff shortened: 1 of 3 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
```diff
@@ -76,24 +81,56 @@ case class RichExecutionObject(exec: Execution.type) {
     * `modifyConf` can be used to update the hive conf that will be used to run the Hive action.
     */
   def fromHive[T](hive: Hive[T], modifyConf: HiveConf => Unit = _ => ()): Execution[T] = {
+    /* Need to get the stack trace outside of the execution so we get a stack trace that is from the
+     * main thread which includes the stack that called this function.
+     */
+    val stacktrace = Thread.currentThread.getStackTrace.tail
```
SamRoberts: Is this expensive?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
```diff
@@ -76,24 +81,56 @@ case class RichExecutionObject(exec: Execution.type) {
     * `modifyConf` can be used to update the hive conf that will be used to run the Hive action.
     */
   def fromHive[T](hive: Hive[T], modifyConf: HiveConf => Unit = _ => ()): Execution[T] = {
+    /* Need to get the stack trace outside of the execution so we get a stack trace that is from the
```
SamRoberts: Did we used to report the stack trace from inside the future's thread? Was there any useful information in there?

stephanh: No. It just provided the stack trace of creating a future within Execution, etc.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
stephanh: Getting the stack trace is somewhat expensive according to Google. However, during the lifetime of the project there should only be a couple of calls of these functions.

I'm not really sure of a better way of doing this.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
```diff
@@ -76,24 +81,56 @@ case class RichExecutionObject(exec: Execution.type) {
...
     }
   }
 
+  /**
+    * Helper function to convert a result to a future.
+    *
+    * For errors it creates a top level exception and fills it in with the provided stack trace.
+    * It also prefixes the exception message with the provided prefix.
+    */
+  def resultToFuture[T](
```
SamRoberts: maybe move `resultToFuture` below all the `fromBlah` functions, so all the from functions are grouped together

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
```diff
@@ -76,24 +81,56 @@ case class RichExecutionObject(exec: Execution.type) {
...
+    Execution.fromFuture[T](_ => resultToFuture(result, "Operation failed", stacktrace))
+  }
+
   /** Changes from an action that produces a scalaz Disjunction to an Execution. */
   def fromEither[T](disjunction: => String \/ T): Execution[T] =
-    Execution.fromFuture(_ => disjunction.fold(
-      msg => Future.failed(new Exception(msg)),
-      x   => Future.successful(x)
-    ))
+    fromResult(disjunction.fold(Result.fail, Result.ok))
```
SamRoberts: shouldn't you use the string error message as the prefix in the `resultToFuture` call?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
```diff
@@ -76,24 +81,56 @@ case class RichExecutionObject(exec: Execution.type) {
...
+    )
+
+  /** Changes from an action that produces a Result to an Execution. */
+  def fromResult[T](result: => Result[T]): Execution[T] = {
+    /* Need to get the stack trace outside of the execution so we get a stack trace that is from the
+     * main thread which includes the stack that called this function.
+     */
+    val stacktrace = Thread.currentThread.getStackTrace.tail
+
+    Execution.fromFuture[T](_ => resultToFuture(result, "Operation failed", stacktrace))
```
SamRoberts: It's a bit annoying to have a generic "Operation failed" message here, when probably there are perfectly good error messages in the result data structure.

SamRoberts:
I don't know, I like the extra stack trace accuracy, but I'm not sure that it will make day to day debugging any easier to add a generic error layer on top of the actual error message, and it's already hard to get all our users to look at the full exception chain.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
stephanh: Will do.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
stephanh: No, we pass the error string to `Result.fail` which will create an Error with that error string. `resultToFuture` will extract that error string and make sure it is included in the exception.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
stephanh: That is only a prefix for the message or if there is only an exception it is the message of the parent exception.

stephanh:
There are 3 error cases:
* **msg** In which case we return an exception with prefix plus message, so the only thing that has changed is that the exception has a more useful stack trace and the message includes a prefix.
* **msg plus exception** We return an exception with prefix plus message and the original exception as cause. Again the only thing that has changed is the more useful stack trace and the prefix as part of the message.
* **exception** This has changed the most. Previously we would return just the original exception. Now we create a new exception so we can get the better stack trace, set the prefix as message and set the original exception as cause.

When debugging I found it quite annoying that I couldn't see at least where in the Execution flow the error occurred.

SamRoberts:
Sure. Can we just change the **exception** case so that we add the original exception message to the new exception message, similar to the other two cases?

stephanh:
But we keep the original exception. It is the cause for the top level exception. Do you want to duplicate the message and have it at the parent level (with the prefix) and the child level without the prefix?

SamRoberts:
Yeah, exactly. I know it's duplicating the message, but I think it's going to make it a lot easier if the top level exception contains the actual error message.

stephanh:
Ok. I'll do that.

SamRoberts:
Awesome. With that change: :+1:

stephanh:
Done.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
stephanh: Done

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
vineethvarghese: Long time back we used to throw an exception catch it there itself to get the stacktrace out. This sure beats that :).

vineethvarghese:
:+1: Looks good. The issues will still persist if we use just Execution.from. I am curious about how other scalding users cope with this.

stephanh:
We should probably discourage the use of `Execution.from` when an error is expected in favour of using `Execution.fromResult`.

rowandavies:
@stephanh Do you mean `Execution.fromFuture`?

PR #343, opened 2015-03-25
splitting uploads

jonjuscorpion666:
Splitting of uploads into 2

1) File Matching Module
2) File Upload Module
3)Backward compactability with current uploads


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -47,6 +40,9 @@ case class UploadInfo(files: List[String]) {
   )
 }
 
+/** Case Class to be exposed and uploaded HDFS */
```
stephanh: Can we make that comment a bit clearer.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
     */
   def upload(config: UploadConfig): Execution[UploadInfo] =
     UploadEx.execution(config)
+  /**
```
stephanh: There should be a new line here.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+   * Get the list of Data files matching the patttern
```
stephanh: `Data` => `data`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+   *
+   * `fileMatching` expects data files intended for HDFS to be placed in the local
+   * ingestion path. `upload` processes all data files that match a given file
+   * pattern. The file pattern format is explained below.
+   *
+   * Some files placed on the local machine are control files. These files
+   * are not intended for HDFS and are ignored by `fileMatching`. `fileMatching` will log
```
stephanh: Mention of HDFS is not relevant here since this doesn't do anything with HDFS.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+   * Some example file patterns:
+   *  - `{table}{yyyyMMdd}.DAT`
+   *  - `{table}_{yyyyMMdd_HHss}.TXT.*.{yyyyMMddHHss}`
+   *  - `??_{table}-{ddMMyy}*`
+   *
+   * @param config: The upload configuration: [[UploadConfig]].
+   *
+   * @return The list of Data files.
+   */
+  def fileMatching(config: UploadConfig): Execution[List[DataFiles]] =
```
stephanh: Can we rename this to `getLocalFiles` or something more meaningful?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+ * any remaining files to HDFS. `upload` will refuse to copy a file if that
+ * file or it's control flags are already present in HDFS. If you
+ * need to overwrite a file that already exists in HDFS, you will need to
+ * delete the file and it's control flags before `upload` will replace it.
+ *
+ * @param config: The upload configuration: [[UploadConfig]].
+ * @param  matchingfile: List of DataFiles matching the pattern: [[DataFiles]]
+ * @return The list of copied hdfs files.
+ */
+  def uploadfiles(config: UploadConfig,matchingfile: List[DataFiles]) =
```
stephanh: Missing space after `,`.

stephanh: We should have the return type annotated here.

stephanh: Change `matchingFile` to just `files`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
-                copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ") }
-                logger.info(s"Upload ended from ${conf.localIngestPath}")
-              }
+      copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ") }
+      logger.info(s"Upload ended")
+    }
   } yield UploadInfo(copied.map(_.dest.toString))
-}
+
+  /** Supporting the legacy upload */
```
stephanh: It's not really legacy at this stage :smile:. Can we move this code into `Upload` instead of having this indirection.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
-                logger.info("Start of upload from ${conf.localIngestPath}")
-                logger.info(s"tableName        = ${conf.tablename}")
-                logger.info(s"filePattern      = ${conf.filePattern}")
-                logger.info(s"controlPattern   = ${conf.controlPattern}")
-                logger.info(s"localIngestPath  = ${conf.localIngestPath}")
-                logger.info(s"localArchivePath = ${conf.localArchivePath}")
-                logger.info(s"hdfsArchivePath  = ${conf.hdfsArchivePath}")
-                logger.info(s"hdfsLandingPath  = ${conf.hdfsLandingPath}")
-              }
+      logger.info("Start of FileMatching from ${conf.localIngestPath}")
```
stephanh: Why did you change the indentation. The previous indentation was our preferred way of aligning things in a `for` comprehension.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -47,6 +40,9 @@ case class UploadInfo(files: List[String]) {
   )
 }
 
+/** Case Class to be exposed and uploaded HDFS */
+case class DataFiles(file: Path, parsedDate: String)
```
SamRoberts: I don't think `Path` is the right type to use for a local file. I also don't think we should essentially duplicate the `Data` type in `Input`. Why not move that type here? I would also prefer the shorter `Data` to `DataFiles`. If we keep the name `DataFiles`, we should make it singular (`DataFile`)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
+    ))
     _      <- Execution.from(
-                files.controlFiles.foreach(ctrl =>logger.info(s"skipping control file ${ctrl.file.getName}"))
-              )
-    copied <- Execution.fromHdfs(
-                Push.push(files.dataFiles, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath)
-              )
+      files.controlFiles.foreach(ctrl =>logger.info(s"skipping control file ${ctrl.file.getName}"))
+    )
+  }yield files.dataFiles.map(_ match {case Data(a, b) =>DataFiles(a.toPath,b) })
```
stephanh: Missing spaces.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -60,7 +56,7 @@ case class UploadInfo(files: List[String]) {
   * @param hdfsLandingPath:  The HDFS landing path where files are copied to.
   * @param localArchivePath: The local archive path where files are archived.
   * @param hdfsArchivePath:  The HDFS archive path, where files are also archived.
-  * @param tableName:        The table name or file name in database or project.
+  * @param tablename:        The table name or file name in database or project.
```
SamRoberts: why change this?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: I would prefer a different name here: these aren't the only matching files we will ever have to deal with: there are also matching control files. Something like `findInputs` seems good to me. The name also needs to be plural.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
stephanh: `.map(_ match { case ...})` => `.map { case ... }`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Or maybe `findSources`, as `source` is the word we have used elsewhere?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+   * @param config: The upload configuration: [[UploadConfig]].
+   *
+   * @return The list of Data files.
+   */
+  def fileMatching(config: UploadConfig): Execution[List[DataFiles]] =
+    UploadEx.matcher(config)
+/**
+ * Pushes source files onto HDFS and archives them locally.
+ *
+ * `upload` expects data files intended for HDFS to be placed in the local
```
SamRoberts: This paragraph isn't relevant to `uploadFiles`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+ * ingestion path. `upload` processes all data files that match a given file
+ * pattern. The file pattern format is explained below.
+ *
+ * Each data file will be copied into the HDFS landing path. Data files are
+ * also compressed and archived on the local machine and on HDFS.
+ *
+ * Some files placed on the local machine are control files. These files
```
SamRoberts: This paragraph isn't relevant to `uploadFiles` either

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+ * are not intended for HDFS and are ignored by `upload`. `upload` will log
+ * a message whenever it ignores a control file.
+ *
+ * When an error occurs, `upload` stops copying files immediately. Once the
```
SamRoberts: Just need to change the `upload` to be `uploadFiles` in this comment

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
stephanh: We should probably curry this as well for easier use within map, etc.

The name should be camel case as well. But can we rename it to `hdfsUpload` or something more creative to make it less similar to `upload`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
-    copied <- Execution.fromHdfs(
-                Push.push(files.dataFiles, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath)
-              )
+      files.controlFiles.foreach(ctrl =>logger.info(s"skipping control file ${ctrl.file.getName}"))
+    )
+  }yield files.dataFiles.map(_ match {case Data(a, b) =>DataFiles(a.toPath,b) })
+
+  /**UploadFilesExecution uploads DataFiles */
+  def upload(conf: UploadConfig,file: List[DataFiles]) = for {
+    _         <- Execution.from {
```
stephanh: Just put that on one line and use `()` instead of `{}` to improve readability.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: And I wonder if, instead of duplicating these comments between upload and these two methods, maybe it would be a good idea to just explain (in `upload`'s comment) that `upload` just runs these two methods in succession?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
-              )
-    copied <- Execution.fromHdfs(
-                Push.push(files.dataFiles, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath)
-              )
+      files.controlFiles.foreach(ctrl =>logger.info(s"skipping control file ${ctrl.file.getName}"))
+    )
+  }yield files.dataFiles.map(_ match {case Data(a, b) =>DataFiles(a.toPath,b) })
+
+  /**UploadFilesExecution uploads DataFiles */
+  def upload(conf: UploadConfig,file: List[DataFiles]) = for {
```
stephanh: Should be files since it is a plural.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
+    _         <- Execution.from {
+      logger.info("Starting fileUploads")
```
SamRoberts: I wonder if there is much point in logging this, when both the child executions log each of their actions

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
+    }
+    datanew   =  file.map(_ match { case  DataFiles(a, b) => Data(a.toFile,b)})
```
stephanh: better variable name?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Oh, sorry, my bad. For a moment I thought this was the upload implementation. Please disregard the above comment.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Is `upload` really legacy? isn't it a valid convenience method if users don't want to inspect the files? Or do we expect people to inspect files 90% of the time?

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -14,9 +14,12 @@
 
 package au.com.cba.omnia.maestro.core.task
 
+import java.nio.file.Paths
+
+import au.com.cba.omnia.maestro.core.task.DataFiles
 import au.com.cba.omnia.thermometer.core.ThermometerSpec
 import au.com.cba.omnia.thermometer.core.Thermometer._
-import au.com.cba.omnia.thermometer.fact.PathFactoids._
+
```
SamRoberts: extra space

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +209,45 @@ trait UploadExecution {
...
+      copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ") }
+      logger.info(s"Upload ended")
+    }
   } yield UploadInfo(copied.map(_.dest.toString))
-}
+
+  /** Supporting the legacy upload */
+  def execution(conf: UploadConfig): Execution[UploadInfo] = for {
+    files      <- matcher(conf)
+    uploadinfo <- upload(conf,files)
```
stephanh: Camel case.

Also in such a simple case just a flatMap might be better:
```
matcher(conf).flatMap(upload(conf, _)))
```

If you use currying it could be even nicer.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +32,8 @@ Upload execution properties
   can custom upload using execution monad                 $custom
   can use continue to stop after an empty upload          $continue
   can upload sequence files for the same timestamp        $sequenceFiles
+  can filematching using execution monad                  $matching
```
SamRoberts: would prefer to use english words in these textual descriptions

stephanh: Can we provide a better test property description here.

version.sbt
```diff
@@ -12,7 +12,7 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-version in ThisBuild := "2.3.2"
+version in ThisBuild := "2.3.3"
```
stephanh: This should probably be a minor version bump and not a patch bump so `2.4.0`.

SamRoberts: Rules for bumping versions:
 - important backwards incompatibility: bump the first number
 - new feature: bump the second number
 - most bug fixes: bump the third number
 - be reasonable when applying the above rules

In this case, you have a new feature users can use, so you should be bumping the second number.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Note that if we avoid having both `Data` and `DataFiles` types, and just keep the one type, then this map can be deleted.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: If we merge `Data` and `DataFiles`, there will be no need for a new variable at all

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
stephanh: I actually prefer tablename :smile:.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: done

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: Thought initially about ```hdfsUpload``` . but felt like its bit confusing as we have feature other than hdfs uploads like local archive etc. Really tough task to find suitable names for functions.

SamRoberts: personally I quite like `uploadFiles`. After all, the only difference between `upload` and `uploadFiles` is that `uploadFiles` explicitly takes a list of files to upload.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: @vineethvarghese .your thoughts please on ```Path``` type.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: I think its a bug in last release . Named wrongly in comments. Actual variable name was ```tablename``` and used ```tableName``` in comments. correct me if i am wrong @SamRoberts

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,26 +14,19 @@
 
 package au.com.cba.omnia.maestro.core.task
 
-import java.io.File
+import java.nio.file.Path
```
vineethvarghese: Why are we using nio Path?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: I thought tableName was the right camel case form.

SamRoberts: @jonjuscorpion666 you are completely correct! I hadn't noticed that ...

Well, in that case, I don't mind whether we use `tablename` or `tableName`, but thanks for noticing the discreprancy.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: what else you are suggesting ?  @vineethvarghese

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: I am not sure about the name DataFiles (`Input` maybe?) and we may not want to use nio Path. Why not just use a string for the path?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: We should be a bit more consistent about our naming. For instance, uploadfiles or uploadFiles. Personally, I prefer the later but I am fine with the former too as long as we are consistent about using that everywhere.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+ * file or it's control flags are already present in HDFS. If you
+ * need to overwrite a file that already exists in HDFS, you will need to
+ * delete the file and it's control flags before `upload` will replace it.
+ *
+ * @param config: The upload configuration: [[UploadConfig]].
+ * @param  matchingfile: List of DataFiles matching the pattern: [[DataFiles]]
+ * @return The list of copied hdfs files.
+ */
+  def uploadfiles(config: UploadConfig,matchingfile: List[DataFiles]) =
+    UploadEx.upload(config,matchingfile)
```
vineethvarghese: Space between arguments.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: Why not just override the toString of UploadConfig to return a String with all this info and just log that string here?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: String :)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: returning string /file make huge difference?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -137,6 +133,73 @@ trait UploadExecution {
...
+ * cause of the error has been addressed, `upload` can be run again to copy
+ * any remaining files to HDFS. `upload` will refuse to copy a file if that
+ * file or it's control flags are already present in HDFS. If you
+ * need to overwrite a file that already exists in HDFS, you will need to
+ * delete the file and it's control flags before `upload` will replace it.
+ *
+ * @param config: The upload configuration: [[UploadConfig]].
+ * @param  matchingfile: List of DataFiles matching the pattern: [[DataFiles]]
+ * @return The list of copied hdfs files.
+ */
```
stephanh: [Codacy](https://www.codacy.com/app/6667/CommBank/file/maestro/issues/source?bid=2205397&fileBranchId=2208061#l201) Public method must have explicit type

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -88,4 +93,42 @@ Upload execution properties
       s"$destDir/2014/11/11"
     ))
   }
+
```
stephanh: [Codacy](https://www.codacy.com/app/6667/CommBank/file/maestro/issues/source?bid=2205397&fileBranchId=2208061#l97) Public method must have explicit type

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -88,4 +93,42 @@ Upload execution properties
...
+    val root      = s"$dir/user"
+    val dirStruct = "normal/mydomain/mytable"
+    val destDir   = "hdfs-root/source/$dirStruct"
+    val conf      = UploadConfig(
+      s"$root/local-ingest/dataFeed/normal/mydomain", destDir,
+      s"$root/local-archive/$dirStruct", "hdfs-root/archive/$dirStruct", "mytable"
+    )
+
+    executesSuccessfully(UploadExec.fileMatching(conf)).map(data =>data.file.toString) must containTheSameElementsAs(List(
+      s"$root/local-ingest/dataFeed/normal/mydomain/mytable_20141111.dat",
```
stephanh: [Codacy](https://www.codacy.com/app/6667/CommBank/file/maestro/issues/source?bid=2205397&fileBranchId=2208061#l108) The string literal /local-ingest/dataFeed/normal/mydomain/mytable_20141112.dat" appears 2 times in the file.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -88,4 +93,42 @@ Upload execution properties
...
+  def matching = withEnvironment(path(getClass.getResource("/upload-execution-normal").toString)) {
+    val root      = s"$dir/user"
+    val dirStruct = "normal/mydomain/mytable"
+    val destDir   = "hdfs-root/source/$dirStruct"
+    val conf      = UploadConfig(
+      s"$root/local-ingest/dataFeed/normal/mydomain", destDir,
+      s"$root/local-archive/$dirStruct", "hdfs-root/archive/$dirStruct", "mytable"
+    )
+
+    executesSuccessfully(UploadExec.fileMatching(conf)).map(data =>data.file.toString) must containTheSameElementsAs(List(
```
stephanh: [Codacy](https://www.codacy.com/app/6667/CommBank/file/maestro/issues/source?bid=2205397&fileBranchId=2208061#l107) The string literal /local-ingest/dataFeed/normal/mydomain/mytable_20141111.dat" appears 2 times in the file.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +32,8 @@ Upload execution properties
   can custom upload using execution monad                 $custom
   can use continue to stop after an empty upload          $continue
   can upload sequence files for the same timestamp        $sequenceFiles
+  can filematching using execution monad                  $matching
+  can uploadfile using execution monad                    $uploadfile
 """
```
stephanh: [Codacy](https://www.codacy.com/app/6667/CommBank/file/maestro/issues/source?bid=2205397&fileBranchId=2208061#l39) The string literal "/upload-execution-normal" appears 3 times in the file.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: Can we add this type of check to codacy?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: @jonjuscorpion666 I believe we discussed this on gitter a couple of days back so you should find the details there. There is no clear right or wrong answer to many things. There might be "rightish/wrongish" answers :-). There are drawbacks to exposing a raw file (at least I think so) but at the same time there are advantages too (in development time mostly).  We will have to weigh both and come up with a pragmatic choice.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: done

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,26 +14,23 @@
...
 import org.apache.log4j.Logger
 
-import org.apache.hadoop.conf.Configuration
-
-import com.twitter.scalding._
+import org.apache.commons.io.input.ReversedLinesFileReader
+import org.apache.commons.io.FileUtils
 
-import au.com.cba.omnia.omnitool.Result
+import java.io._
```
vineethvarghese: Do we need a wild card import here?

vineethvarghese: Also this doesn't follow our convention of import listing

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,7 +75,18 @@ case class UploadConfig(
   tablename: String,
   filePattern: String = "{table}?{yyyyMMdd}*",
   controlPattern: Regex = ControlPattern.default
-)
+){
+  override def toString=
+    s"""TableName        =${tablename},
+       |filePattern      =${filePattern},
```
vineethvarghese: Is that 2 space or 3 space? Maybe a space after = too

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
     *
     * @param config: The upload configuration: [[UploadConfig]].
     *
-    * @return The list of copied hdfs files.
+    * @return UploadInfo: Upload info
     */
-  def upload(config: UploadConfig): Execution[UploadInfo] =
-    UploadEx.execution(config)
+  def upload(config: UploadConfig): Execution[UploadInfo] ={
```
vineethvarghese: Space after =

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+    for {
+      uploadDetails <- UploadEx.matcher(config).flatMap(UploadEx.upload(config,_))
```
vineethvarghese: Do we need a for expression here? If you are using a for expression then we should use it all the way. Why have one flatMap call here?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Please don't use wildcard imports unless you're using a large number of classes from the package

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,26 +14,23 @@
...
-import au.com.cba.omnia.omnitool.Result
+import java.io._
 
-import au.com.cba.omnia.permafrost.hdfs.Hdfs
+import com.twitter.scalding._
 
 import au.com.cba.omnia.maestro.core.scalding.ConfHelper
 import au.com.cba.omnia.maestro.core.scalding.ExecutionOps._
-import au.com.cba.omnia.maestro.core.upload.{ControlPattern, Copied, Input, Push}
+import au.com.cba.omnia.maestro.core.upload._
```
SamRoberts: likewise: how many things are you importing from upload? I would prefer the deleted line to this one unless you are importing perhaps another 2-3 things?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
+package au.com.cba.omnia.maestro.core.upload
+
+import java.io.{FileReader, BufferedReader}
+
+import org.apache.commons.io.FileUtils
+import org.apache.commons.io.input.ReversedLinesFileReader
+
+/* To Parse Trailer information */
```
vineethvarghese: Isn't this the trailer information itself? Does it have anything to do parsing?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+case class Trailer(rtype: String, count: Int, checksum: Int, checksumColumn: String)
+
+/* To Parse Header information */
```
vineethvarghese: Same comment as above

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,26 +14,23 @@
...
+import au.com.cba.omnia.omnitool.Result
```
SamRoberts: Minor: our import ordering rules essentially say "put the most generic imports first, followed by the most specific". In this case, you have the more generic omnitool import following the very specific maestro imports. I would prefer it if the omnitool import was above the maestro import.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -47,6 +44,9 @@ case class UploadInfo(files: List[String]) {
   )
 }
 
+/** Case Class to be exposed for getting the file details and used in uploads */
+case class Data(file: String, fileSubDir: String)
```
SamRoberts: Previously, in your `DataFiles` type, you had changed the name of `fileSubDir` to be something like `fileDate` (can't remember the exact name)?

I presume you did that because you found it easier to understand. Maybe you should rename the field in `Data` too?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+case class Header(rtype: String, businessDate: String, extractTime: String, fileName: String)
```
vineethvarghese: Why is business date string a String here but in the trailer we convert count to int

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
+package au.com.cba.omnia.maestro.core.upload
+
+import java.io.{FileReader, BufferedReader}
+
+import org.apache.commons.io.FileUtils
+import org.apache.commons.io.input.ReversedLinesFileReader
+
+/* To Parse Trailer information */
+case class Trailer(rtype: String, count: Int, checksum: Int, checksumColumn: String)
```
vineethvarghese: Shouldn't we be using Long here?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,7 +75,18 @@ case class UploadConfig(
   tablename: String,
   filePattern: String = "{table}?{yyyyMMdd}*",
   controlPattern: Regex = ControlPattern.default
-)
+){
+  override def toString=
```
SamRoberts: Hrmm ... I'm a little on the fence about this change. On the one hand, this will make `UploadConfig.toString` a lot prettier. On the other hand, it's inconsistent with the way we do every other `FooConfig.toString`, and it means that you can't put the `UploadConfig.toString` inside some other structure and call `toString` on that and get anything sensible out.

Actually, thinking about it a bit more, I'm against changing `toString`. By all means have some sort of `prettyPrintConfig` method (or whatever you want to call it), though. That would be a nice improvement over the previous code.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,7 +75,18 @@ case class UploadConfig(
...
+    s"""TableName        =${tablename},
```
SamRoberts: we should have a space between the equals and the values of each parameter, I think.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
     *
     * @param config: The upload configuration: [[UploadConfig]].
     *
-    * @return The list of copied hdfs files.
+    * @return UploadInfo: Upload info
```
SamRoberts: I understand why you changed this, but it's a bit of a tautology. Of course the `UploadInfo` parameter is the `Upload info`! Please expand the description to explain what is contained in `UploadInfo`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: To be fair, I did ask @jonjuscorpion666 to try this approach :).

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Either:

```
for {
  files <- UploadEx.matcher(config)
  info <- UploadEx.upload(config, files)
} yield info
```

or

```
UploadEx.matcher(config).flatMap(UploadEx.upload(config, _))
```

Whenever you see something like this:
```
for {
  foo <- bar
} yield foo
```

That is equivalent to just: `bar`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -47,6 +44,9 @@ case class UploadInfo(files: List[String]) {
   )
 }
 
+/** Case Class to be exposed for getting the file details and used in uploads */
```
vineethvarghese: How about something like - `Data` is a representation of etc..etc. rather than saying it is a case class to be exposed and all that.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+
+/* To Parse Header information */
+case class Header(rtype: String, businessDate: String, extractTime: String, fileName: String)
+
+/*Extract the header & Trailer from the file */
+case class headerTrailer(header: String, trailer: String)
```
vineethvarghese: The obvious thing is that the class name should start with a capital letter but I am still not clear why this class does. Maybe if I read more I will see.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object ParserHT {
+
+  /** splitting header and returning a Header case class */
```
vineethvarghese: Maybe just - Splitting the header and returning the contents as a `Header`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
```diff
@@ -29,6 +29,8 @@ import au.com.cba.omnia.omnitool.file.ops.Temp
 
 import au.com.cba.omnia.permafrost.hdfs.Hdfs
 
+import au.com.cba.omnia.maestro.core.task.Data
```
vineethvarghese: We have a cyclic package dependency here.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+  }
+
+  /**
+   * Get the data files matching the pattern
+   *
+   * See the description of the file pattern at [[au.com.cba.omnia.maestro.core.task.UploadExecution.upload()]]
+   *
+   * @param config: The upload configuration: [[UploadConfig]].
+   *
+   * @return List[Data]  The list of Data files.
```
SamRoberts: `Data` -> `data`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   */
+  def findSources(config: UploadConfig): Execution[List[Data]] =
```
vineethvarghese: I am in 2 minds about this but maybe findData?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   *
+   * @param header   : file trailer
+   * @return Header : returns Trailer case class
+   */
+
+  def parseHeader[A] (header: String): Header =
+    ParserHT.parseH(header)
+
+  /**
+   *  Trailer Parser function splits and parse the delimited file trailer string
```
SamRoberts: If I remember correctly, throughout the rest of this file we have the vertical line of asterisks in line with the second asterisk, not the first.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+ *
+ * @param config  : The upload configuration: [[UploadConfig]]
+ * @param files  : List of Data matching the pattern: [[Data]]
+ * @return UploadInfo :The list of copied hdfs files : Execution[UploadInfo]
+ */
+  def uploadFiles(config: UploadConfig) (files: List[Data]): Execution[UploadInfo] =
+    UploadEx.upload(config, files)
+
+  /**
+   *  Header Parser function splits and parse the delimited file header string
```
SamRoberts: Please redo comment to have a summary line at the top, then a paragraph of explanatory text below it, as per the other scaladoc comments in this file.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Also, please redo comments to have the summary line and then the rest of the explanatory test below it, as per other comments in this file.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   *  into Header case class.User can write custom function if default
+   *  function dont satisfy the requirment ie if file header does satisfy
```
SamRoberts: how does the user know what the default pattern is?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   * @return Header : returns Trailer case class
+   */
+
+  def parseHeader[A] (header: String): Header =
+    ParserHT.parseH(header)
+
+  /**
+   *  Trailer Parser function splits and parse the delimited file trailer string
+   *  into Trailer case class.User can write custom function if default
+   *  function dont satisfy the requirment ie if file trailer does satisfy
```
SamRoberts: how does the user know what the default pattern is?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   *  the default pattern.
+   *
+   * @param trailer : file trailer
+   * @return Trailer: returns Trailer case class
+   */
+  def parseTrailer[B] (trailer: String): Trailer =
+    ParserHT.parseT(trailer)
+
+  /**
+   * Extract the header and trailer from the file and returns
```
SamRoberts: This is a great scaladoc :+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   *  parsed value of header and trailer.
+   *
+   * Default parser function is designed on with assumption that header and
+   * trailer matches below pattern.
+   *
+   * H|BusinessDate|ExtractTime|FileName
+   * T|RecordCount|CheckSumValue|CheckSumColumn
+   *
+   * User can write custom function if custom function don't satisfy the requirment
```
SamRoberts: `custom function if custom function don't satisfy the requirment` -> `their own custom function if the default function doesn't satisfy their requirements`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   * @return Trailer: returns Trailer case class
+   */
+  def parseTrailer[B] (trailer: String): Trailer =
+    ParserHT.parseT(trailer)
+
+  /**
+   * Extract the header and trailer from the file and returns
+   *  parsed value of header and trailer.
+   *
+   * Default parser function is designed on with assumption that header and
```
SamRoberts: `on with` -> `on` or `with`. Either word is sufficient

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   * trailer matches below pattern.
+   *
+   * H|BusinessDate|ExtractTime|FileName
+   * T|RecordCount|CheckSumValue|CheckSumColumn
+   *
+   * User can write custom function if custom function don't satisfy the requirment
+   *
+   * @param parseH          : function that parse the header
```
SamRoberts: `function that parse` -> `function to parse` or `function that parses`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   * @param parseT          : function that parse the trailer
+   * @return Result[(H, T)] : returns a result monad of header and trailer function
```
SamRoberts: As a general rule of thumb, when we return some monadic computation returning a value, our scaladoc tends to just describe the value, not the monadic computation.

Also, this sounds like you are returning the header and trailer functions themselves, whereas you are actually returning the result of running the header and trailer functions.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +228,42 @@ trait UploadExecution {
...
-    files  <- Execution.fromEither(Input.findFiles(
-                conf.localIngestPath, conf.tablename, conf.filePattern, conf.controlPattern
-              ))
-    _      <- Execution.from(
-                files.controlFiles.foreach(ctrl =>logger.info(s"skipping control file ${ctrl.file.getName}"))
-              )
+  /** fileMatchingExecution to extract the datafiles matched with the filePattern */
+  def matcher(conf: UploadConfig): Execution[List[Data]] = for {
+    _ <- Execution.from {
+      logger.info("Start of FileMatching from ${conf.localIngestPath}")
```
SamRoberts: `FileMatching` -> some english phrase

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +228,42 @@ trait UploadExecution {
...
+    files <- Execution.fromEither(Input.findFiles(
+      conf.localIngestPath, conf.tablename, conf.filePattern, conf.controlPattern
+    ))
+    _ <- Execution.from(
+      files.controlFiles.foreach(ctrl => logger.info(s"skipping control file ${ctrl.file.getName}"))
+    )
+  } yield files.dataFiles
+
+  /** UploadFilesExecution uploads Data */
+  def upload(conf: UploadConfig, files: List[Data]): Execution[UploadInfo] = for {
```
SamRoberts: I don't think we should call this upload. We already have an `upload` method, and it does both the finding files, and the uploading them to HDFS. It's a bit confusing to reuse the same word to mean something different here.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +228,42 @@ trait UploadExecution {
...
+    )
+    _ <- Execution.from {
+      copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ")}
+      logger.info(s"Upload ended")
+    }
   } yield UploadInfo(copied.map(_.dest.toString))
-}
+
+  /** Extract Header and trailer and file and Parse same*/
+  def parser[H, T](parseH: String => H, parseT: String => T, file: String): Result[(H, T)]= {
```
SamRoberts: We have a file parser concept as well. We should probably be a little more specific with this name. `htParser`? I'm not sure ...

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +228,42 @@ trait UploadExecution {
...
+    val inputReverse       = new ReversedLinesFileReader(FileUtils.getFile(file))
+    val inputHeader        = new BufferedReader(new FileReader(file))
+    val trailer            = Result.ok(inputReverse.readLine())
```
SamRoberts: What if there are more than one H or T line? That should be an error too, right?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: So we have a parser function here, and a `ParserHT` object below as well? Why have we split up the underlying logic into two places?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   * H|BusinessDate|ExtractTime|FileName
+   * T|RecordCount|CheckSumValue|CheckSumColumn
+   *
+   * User can write custom function if custom function don't satisfy the requirment
+   *
+   * @param parseH          : function that parse the header
+   * @param parseT          : function that parse the trailer
+   * @return Result[(H, T)] : returns a result monad of header and trailer function
+   */
+  def parseHT[H, T](parseH: String => H, parseT: String => T, file: String): Result[(H, T)] =
```
SamRoberts: If we have default functions, maybe we should actually make them the default values to put here?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Nevermind, I see why the split. I have a few more comments about that, but they will go on other lines. My point about the name `parser` remains.

SamRoberts: Why doesn't the type of the parse functions include the possibility that the parse will fail?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: What happens if the file is empty?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +228,42 @@ trait UploadExecution {
...
-}
+
+  /** Extract Header and trailer and file and Parse same*/
+  def parser[H, T](parseH: String => H, parseT: String => T, file: String): Result[(H, T)]= {
+    val inputReverse       = new ReversedLinesFileReader(FileUtils.getFile(file))
+    val inputHeader        = new BufferedReader(new FileReader(file))
+    val trailer            = Result.ok(inputReverse.readLine())
+    val header             = Result.ok(inputHeader.readLine())
+    for{
+      head    <-  header.map(parseH)
```
SamRoberts: What happens of the parse function fails?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +228,42 @@ trait UploadExecution {
...
+      trail   <-  trailer.map(parseT)
```
SamRoberts: What if the user just wants to retrieve the header or trailer individually, not both at the same time?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: We don't actually ever use this, do we?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+/* To Parse Trailer information */
+case class Trailer(rtype: String, count: Int, checksum: Int, checksumColumn: String)
+
+/* To Parse Header information */
+case class Header(rtype: String, businessDate: String, extractTime: String, fileName: String)
+
+/*Extract the header & Trailer from the file */
+case class headerTrailer(header: String, trailer: String)
+
+object ParserHT {
```
SamRoberts: So it seems we see this object as a private API, and are exposing the methods through the `Upload` trait? I don't see much point in that. In general, when we want to expose a bunch of constants to the user, what we have done is have a public object containing those constants. Can we make this object part of the public API, and remove the corresponding methods from the `Upload`?

In order to avoid the user having to write `ParserHT.parseH` all the time, you can make these functions the default value for the `parser` method, and that's even more convenient.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  /** splitting header and returning a Header case class */
+  def parseH(header: String): Header = {
+    val headerSplit = header.split('|')
+    Header(headerSplit(0), headerSplit(1), headerSplit(2), headerSplit(3))
```
SamRoberts: What if the header doesn't have the right format?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  /** splitting trailer and returning a Trailer case class */
+  def parseT(trailer: String): Trailer = {
+    val trailerSplit = trailer.split('|')
+    Trailer(trailerSplit(0), trailerSplit(1).toInt, trailerSplit(2).toInt, trailerSplit(3))
```
SamRoberts: And likewise: what if the format of the trailer is wrong?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
```diff
@@ -54,7 +56,7 @@ object Push {
     val fileGroups = files.groupBy(_.fileSubDir).mapValues(subDirFiles => subDirFiles.map(_.file)).toList
 
     fileGroups.traverse[Hdfs, Copied] { case (subDir, sources) =>
-      pushToSubDir(sources, subDir, destRoot, archRoot, hdfsArchRoot)
+      pushToSubDir(sources.map(source => new File(source)), subDir, destRoot, archRoot, hdfsArchRoot)
```
SamRoberts: Instead of converting to and from `File` all the time, maybe we should remove `File` from the other apis too.

@vineethvarghese I'm still not sure why you are recommending to avoid `File` in public interfaces? isn't it even more important to use specific types in public interfaces? Are you thinking that we would want to change away from `File` in the future, and want to use the "lowest common denominator" type to help maintain backwards compatibility?

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
SamRoberts: These messages are auto-generated from codacy. Please take them on their merits. We don't normally worry about putting explicit types in all our test methods (which isn't to say it would be a bad thing to do!).

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: Why does the trailer class include the "T"?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: Why does the header class include the "H"?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: Don't you want to check that the header starts with "H"?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: Don't you want to check that the trailer starts with "T"?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+case class Header(rtype: String, businessDate: String, extractTime: String, fileName: String)
+
+/*Extract the header & Trailer from the file */
+case class headerTrailer(header: String, trailer: String)
+
+object ParserHT {
+
+  /** splitting header and returning a Header case class */
+  def parseH(header: String): Header = {
+    val headerSplit = header.split('|')
```
SamRoberts: Shouldn't we use maestro config's splitter value to split the header?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,31 @@
...
+    Header(headerSplit(0), headerSplit(1), headerSplit(2), headerSplit(3))
+  }
+
+  /** splitting trailer and returning a Trailer case class */
+  def parseT(trailer: String): Trailer = {
+    val trailerSplit = trailer.split('|')
```
SamRoberts: Shouldn't we use maestro config's splitter value to split the trailer?

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/ParserHTSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+
+Header Trailer Extracting properties
+===============
+
+
+  split the header and return Header case class   $splitheader
+  split the trailer and return Trailer case class $splittrailer
+"""
+
+  def splitheader  = ParserHT.parseH("H|20150311|20150318131001|hls_20150301.dat") must_== (Header("H","20150311","20150318131001","hls_20150301.dat"))
```
SamRoberts: Have you user used scalacheck before? This would be the perfect place for some scalacheck properties, if you are interested in learning about them

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/ParserHTSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+
+class ParserHTSpec extends ThermometerSpec {
+  def is = s2"""
+
+Header Trailer Extracting properties
+===============
+
+
+  split the header and return Header case class   $splitheader
+  split the trailer and return Trailer case class $splittrailer
```
SamRoberts: We would normally test failure cases as well as success cases

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: `sources` is the word we have used elsewhere (e.g., it's what our default directory structure says, it's what is in `UploadInfo`, etc.)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+ * @param config  : The upload configuration: [[UploadConfig]]
+ * @param files  : List of Data matching the pattern: [[Data]]
+ * @return UploadInfo :The list of copied hdfs files : Execution[UploadInfo]
+ */
+  def uploadFiles(config: UploadConfig) (files: List[Data]): Execution[UploadInfo] =
+    UploadEx.upload(config, files)
+
+  /**
+   *  Header Parser function splits and parse the delimited file header string
+   *  into Header case class.User can write custom function if default
```
vineethvarghese: `into Header case class.User can` -> `into Header case class. User can`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -133,10 +144,81 @@ trait UploadExecution {
...
+   *  function dont satisfy the requirment ie if file header does satisfy
+   *  the default pattern.
+   *
+   * @param header   : file trailer
+   * @return Header : returns Trailer case class
+   */
+
+  def parseHeader[A] (header: String): Header =
```
vineethvarghese: Have we considered exposing the ParserHT object rather than hiding them via these methods?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
vineethvarghese: I think what @SamRoberts is asking is why do we need the rtype in Header/Trailer after forming the Header/Trailer type?

jonjuscorpion666: @ i assume your question is  why we need to parse "H" if header is H|2-03-2014|03:44:00pm|abc .

i thought of adding same because user can validate the parse line is really header or not .

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: I approve of refactoring those log statements into the one coherent method. Just not making that the `toString` method. There is always some tension between the different uses of `toString`, but I think it's safe to say that `toString` should be able to be called in the middle of some other string and not wreck the formatting of that string.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
vineethvarghese: "Are you thinking that we would want to change away from File in the future, and want to use the "lowest common denominator" type to help maintain backwards compatibility?" - yes. Although we could argue that we can make back incompatible changes when required

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: I think our parsing function should be validating whether the parse line is really header or not. In the rare case that's not appropriate, the user can supply their own custom function.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
SamRoberts: Why would we want to move away from File? Is that because of the pending `nio` package? (Which I haven't looked at)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: Do we want to fail if there are more than four fields? I would suggest we do.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: Do we want to fail if there are more than four fields? I would suggest we do.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: i am importing about 7 thing from uploads .

SamRoberts: Ok, fair enough

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
vineethvarghese: Do we gain a lot by exposing a file? 

How about?
`pushToSubDir(sources.map(source => new File(source)), subDir, destRoot, archRoot, hdfsArchRoot)` -> `pushToSubDir(sources.map(new File(_)), subDir, destRoot, archRoot, hdfsArchRoot)`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: How we do something like 

findSource
uploadSource
or
findData
uploadData

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
jonjuscorpion666: @vineethvarghese  . We dont have much info on date format in each files. Hence thought of using String.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: findSources
uploadSources

I think sources is the "most established" name for them within upload (we haven't really been completely consistent, which is my fault). The should be plural though.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
SamRoberts: Well, I guess in general the idea behind strongly typed code is to give specific types to things. We want to prevent people from accidentally using the db password as the file: things like that.

I am certainly open to the suggestion that we want to use a lowest common denominator in this case because the `nio` package gives us a reason to believe `File` will soon be outdated. I only thought of that possibility during this conversation.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: Its mentioned under Parse HT  . i will give a reference link here too

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
vineethvarghese: **Nothing** gets outdated in the java api :). So, by that token we could continue using File. At a later point when we switch to copying with nio, we could convert those file paths to nio.Path.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: @SamRoberts  reg mutiple header and trailer scenario : I am not really sure  we need to really handle  .If we want to find mutiple headers/trailers ,we need  parsing through the file to find all headers and trailers,because second trailer can come anywhere in the file.here we are picking only first and last row.
We are assuming -(from existing patterns) first row of the file is header and last row of the file is trailer.
 And infact If we have mutiple header/Trailer count validation will fail.

SamRoberts: Good points, @jonjuscorpion666. Ok then.

stephanh:
@SamRoberts I assume this PR is superseded by #351 as well now?

SamRoberts:
yes, this was superseded by #351, thanks for picking that up

PR #347, opened 2015-03-29
Added comment on complex thrift types.

rowandavies:


README.md
```diff
@@ -162,7 +162,11 @@ _know_ will work before we run the code (for a valid schema).
 | i32: A 32-bit signed integer                      | INT (4-byte signed integer, from -2,147,483,648 to 2,147,483,647)                             | int           |
 | i64: A 64-bit signed integer                      | BIGINT (8-byte signed integer, from -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807)  | BigInteger    |
 | double: A 64-bit floating point number            | DOUBLE (8-byte double precision floating point number)                                        | double        |
-| string: Encoding agnostic text or binary string   | string 
+| string: Encoding agnostic text or binary string   | string | String
```
SamRoberts: I know it's not your words, but while we're here ... `string` doesn't really work for binary, does it? I believe the string is encoded as UTF8.

rowandavies: Ah - you mean a binary string can't be directly stored as a hive string?  I'll check the code and/or test it.

SamRoberts: I'm pretty sure @laurencer has run into this problem.

SamRoberts:
:+1: 

(I know I have an outstanding issue above, but that's not directly related to your change. It would be great if you follow the issue up, but your change itself is perfectly fine.)

rowandavies:
Actually, I think I should qualify the statement that complex thrift types aren't supported by maestro.  There's no direct support for them, but there is in the supporting libraries, and you can manually decode/encode string fields to complex types by mapping over the rows.

SamRoberts:
Fair enough

SamRoberts:
:+1:

PR #350, opened 2015-03-30
Header Trailer Validatation and Upload split

jonjuscorpion666:
Updated code for Header Trailer Validatation and Upload Split .


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,59 +14,61 @@
 
 package au.com.cba.omnia.maestro.core.task
 
-import java.io.File
-
-import scala.concurrent.Future
 import scala.util.matching.Regex
 
+import java.io.BufferedReader
```
SamRoberts: Minor: `import java.io.{BufferedReader, FileReader}`

Unless you think this looks better as it is

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,59 +14,61 @@
...
   /** Returns a non-emtpy list of sources if we should continue the load.
-    * Fails otherwise. */
+    * Fails otherwise.
+    */
   def withSources: Execution[List[String]] = Execution.from(
     if (continue) files else throw new Exception("No files to upload")
   )
 }
 
+/** Data - A representation the file file to be uploaded and business date */
```
vineethvarghese: 2 Files?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *  - `{table}{yyyyMMdd}.DAT`
+   *  - `{table}_{yyyyMMdd_HHss}.TXT.*.{yyyyMMddHHss}`
+   *  - `??_{table}-{ddMMyy}*`
+   *
+   * @param config: The upload configuration: [[UploadConfig]].
+   *
+   * @return UploadInfo: Returns a non-emtpy list of sources if we should continue the load
+   */
+  def upload(config: UploadConfig): Execution[UploadInfo] = {
+    UploadEx.matcher(config).flatMap(UploadEx.uploader(config,_))
```
vineethvarghese: Space between `config` and `_`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+  }
+
+  /**
+   *   Extract the header from the file and returns header data (Default Implementation)
```
vineethvarghese: Extra indentation?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
   tablename: String,
   filePattern: String = "{table}?{yyyyMMdd}*",
   controlPattern: Regex = ControlPattern.default
-)
+ ){
+  def printUploadConfig=
```
SamRoberts: seeing as this is a method on `UploadConfig`, there is no need for the `UploadConfig` to be repeated in the method name. `print` or `prettyPrint` would both be fine

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   */
+  def upload(config: UploadConfig): Execution[UploadInfo] = {
+    UploadEx.matcher(config).flatMap(UploadEx.uploader(config,_))
+  }
+
+  /**
+   *   Extract the header from the file and returns header data (Default Implementation)
+   *
+   *  Header Parser function splits and parse the delimited file header string
+   *  into Header case class. Default Parser is implemented with assumption that
```
vineethvarghese: Can we rephrase this line? What Header case class are we talking about here? Isn't it just type `H` :)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *  Header follow pattern specified in  [[au.com.cba.omnia.maestro.core.task.UploadExecution.parseHT()]]
+   *  User can write custom function if default
+   *  function doesn't satisfy their requirment ie if file header is not in
+   *  the default pattern.
+   *
+   * @param parseH   : function to parse Header
+   * @param file     : input sourse file
```
vineethvarghese: sourse -> source

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   * @return         : returns file Header data
```
vineethvarghese: returns header type `H` ?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   */
+  def parseHeader[H] (parseH: String => Result[H] = ParserHT.parseH(_)) (file: String): Result[H] =
+    UploadEx.hParser(parseH, file)
+
+  /**
+   *  Extract the trailer from the file and returns (Default Implementation) (Default Implementation)
```
vineethvarghese: Why are we saying "(Default Implementation)" here?. Shouldn't the default just be mentioned somewhere in the details.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *      - second of minute (s).
+   *
+   * Some example file patterns:
+   *  - `{table}{yyyyMMdd}.DAT`
+   *  - `{table}_{yyyyMMdd_HHss}.TXT.*.{yyyyMMddHHss}`
+   *  - `??_{table}-{ddMMyy}*`
+   *
+   * @param config: The upload configuration: [[UploadConfig]].
+   *
+   * @return UploadInfo: Returns a non-emtpy list of sources if we should continue the load
```
SamRoberts: `Returns a` -> `a` (we already know this is the return value)
`non-emtpy` -> `non-empty` (spelling)

But wait, the comments about non-empty aren't true, are they? Upload doesn't automatically fail if there are no files, or at least, it didn't used to. (But maybe it should.)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *  Header follow pattern specified in  [[au.com.cba.omnia.maestro.core.task.UploadExecution.parseHT()]]
+   *  User can write custom function if default
+   *  function doesn't satisfy their requirment ie if file header is not in
+   *  the default pattern.
+   *
+   * @param parseH   : function to parse Header
+   * @param file     : input sourse file
+   * @return         : returns file Header data
+   */
+  def parseHeader[H] (parseH: String => Result[H] = ParserHT.parseH(_)) (file: String): Result[H] =
```
SamRoberts: Good to see you returning a `Result` from the parser function :smile: 

But I'm still not convinced we want this method at all. Why not expose the `ParserHT` object and use default parameters as suggested in your last PR?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    UploadEx.hParser(parseH, file)
+
+  /**
+   *  Extract the trailer from the file and returns (Default Implementation) (Default Implementation)
+   *
+   *  Trailer Parser function splits and parse the delimited file trailer string
+   *  into Trailer case class. Parser is implemented with assumption that
```
vineethvarghese: Again there is no Trailer case class here. Maybe we can just say that we use the user provide function to parse the trailer from the file or something. Of course we have to mentioned the default parser we are using.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   * @return UploadInfo: Returns a non-emtpy list of sources if we should continue the load
+   */
+  def upload(config: UploadConfig): Execution[UploadInfo] = {
+    UploadEx.matcher(config).flatMap(UploadEx.uploader(config,_))
+  }
+
+  /**
+   *   Extract the header from the file and returns header data (Default Implementation)
+   *
+   *  Header Parser function splits and parse the delimited file header string
```
SamRoberts: If we keep this method, and I'm not sure we should, then you will just need to reword the paragraph slightly.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *  Header follow pattern specified in  [[au.com.cba.omnia.maestro.core.task.UploadExecution.parseHT()]]
+   *  User can write custom function if default
+   *  function doesn't satisfy their requirment ie if file trailer is not in
+   *  the default pattern.
+   *
+   * @param parseT : function to parse Trailer
+   * @param file  : input source file
+   * @return      : returns file Trailer information
+   */
+  def parseTrailer[T] (parseT: String => Result[T] = ParserHT.parseT(_)) (file: String): Result[T] =
```
SamRoberts: My comments on `parseHeader` apply here too

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    UploadEx.matcher(config)
+
+  /**
+   * Pushes source files onto HDFS and archives them locally.
+   *
+   * See the description of the HDFS push and archives at [[au.com.cba.omnia.maestro.core.task.UploadExecution.upload()]]
+   *
+   * @param config  : The upload configuration: [[UploadConfig]]
+   * @param files  : List of Data matching the pattern: [[Data]]
+   * @return UploadInfo :The list of copied hdfs files : Execution[UploadInfo]
```
vineethvarghese: You have left space between params and returns in other methods but not here.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    UploadEx.tParser(parseT, file)
+
+  /**
+   * Get the data files matching the pattern
+   *
+   * See the description of the file pattern at [[au.com.cba.omnia.maestro.core.task.UploadExecution.upload()]]
+   *
+   * @param config: The upload configuration: [[UploadConfig]].
+   *
+   * @return List[data]  The list of data files.
```
SamRoberts: Why are you including the return type in the documentation? Is that some scaladoc feature? Shouldn't scaladoc already know the return type and automatically display it to the user?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *
+   * H|BusinessDate|ExtractTime|FileName
+   * T|RecordCount|CheckSumValue|CheckSumColumn
+   *
+   * User can write custom function if custom function doesn't satisfy the requirment
+   *
+   * @param parseH                  : function that parses the header
+   * @param parseT                  : function that parses the trailer
+   * @return (Header,Trailer)       :  parsed header and trailer values
+   */
```
vineethvarghese: Nicely said :). I wish the other parse function follows this pattern.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+  def parseHT[H, T](parseH: String => Result[H] = ParserHT.parseH(_))(parseT: String => Result[T] = ParserHT.parseT(_)) (file: String): Result[(H,T)] = {
```
vineethvarghese: Why do we have a space before the last argument i.e. `(file: String)`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    for{
```
vineethvarghese: Space after for

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+      header   <- UploadEx.hParser(parseH,file)
```
vineethvarghese: Space between arguments

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
-              ))
-    _      <- Execution.from(
-                files.controlFiles.foreach(ctrl =>logger.info(s"skipping control file ${ctrl.file.getName}"))
-              )
+  /** fileMatchingExecution to extract the datafiles matched with the filePattern */
+  def matcher(conf: UploadConfig): Execution[List[Data]] = for {
+    _ <- Execution.from {
+      logger.info("Start of pattern matching from ${conf.localIngestPath}")
+      logger.info(conf.printUploadConfig)
+    }
```
vineethvarghese: Bracket should match E in Execution

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: This function is exposed if user want to extract only Header/Trailer and not together.  Here functionality is different  as compared to previous PR. This is modified as per your suggestion.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    )
+    _ <- Execution.from {
+      copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ")}
+      logger.info(s"Upload ended")
+    }
   } yield UploadInfo(copied.map(_.dest.toString))
+
+  /** Extract Header from file and Parse */
+  def hParser[H](parseH: String => Result[H], file: String): Result[H] = {
+    val inputHeader = new BufferedReader(new FileReader(file))
```
vineethvarghese: What will happen if an exception is thrown here?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    val header = Result.ok(inputHeader.readLine())
+    header.flatMap(parseH)
+  }
+
+  /** Extract Trailer from file and Parse */
+  def tParser[T](parseT: String => Result[T], file: String): Result[T] = {
+    val inputReverse = new ReversedLinesFileReader(FileUtils.getFile(file))
```
vineethvarghese: Same comment here. Operations in file are fundamentally unsafe, so it could fail with an exception here.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    val trailer = Result.ok(inputReverse.readLine())
```
vineethvarghese: If readLine thrown an exception, Result.ok won't deal with it. Try using Result.safe instead.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Ah, I see. Apologies. Will take another look.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
+package au.com.cba.omnia.maestro.core.upload
+
+import java.io.{FileReader, BufferedReader}
+
+import org.apache.commons.io.input.ReversedLinesFileReader
+
+import au.com.cba.omnia.omnitool.Result
+import org.apache.commons.io.FileUtils
```
vineethvarghese: Shouldn't omnia packages come below everything else?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
...
+
+/* Header information - RowType, BusinessDate, ExtractTime, FileName */
+case class Header(businessDate: String, extractTime: String, fileName: String)
+
+object ParserHT {
+
+  /** splitting header and returning content as a Header */
+  def parseH(header: String): Result[Header] = {
+    val headerSplit = Splitter.delimited("|").run(header)
+    if(headerSplit(0).trim != "H" || headerSplit.length != 4)
```
vineethvarghese: Can the headerSplit(0) fail? I am not sure but thought I would ask.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Can you explain the reason for using two parameter lists instead of one? I can think of some decent reasons, but I'm curious what your reasons were.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   * @param config  : The upload configuration: [[UploadConfig]]
+   * @param files  : List of Data matching the pattern: [[Data]]
+   * @return UploadInfo :The list of copied hdfs files : Execution[UploadInfo]
+   */
+  def uploadSources(config: UploadConfig) (files: List[Data]): Execution[UploadInfo] =
+    UploadEx.uploader(config, files)
+
+  /**
+   * Extract the header and trailer from the file and returns
+   *  parsed value of header and trailer.
```
SamRoberts: extra space between `*` and `parsed`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
vineethvarghese: I generally put positives check in if (if I can) because I think it is more readable but this is my personal choice.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *
+   * Default parser function is designed with assumption that header and
+   * trailer matches below pattern.
+   *
+   * H|BusinessDate|ExtractTime|FileName
+   * T|RecordCount|CheckSumValue|CheckSumColumn
+   *
+   * User can write custom function if custom function doesn't satisfy the requirment
```
SamRoberts: Why would user write a custom function if their custom function doesn't satisfy their requirement? :smile:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Agreed :+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: Good catch :)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Push.scala
```diff
@@ -51,10 +53,10 @@ object Push {
     * source or multiple destination files while this method runs.
     */
   def push(files: List[Data], destRoot: String, archRoot: String, hdfsArchRoot: String): Hdfs[List[Copied]] = {
-    val fileGroups = files.groupBy(_.fileSubDir).mapValues(subDirFiles => subDirFiles.map(_.file)).toList
+    val fileGroups = files.groupBy(_.parsedDate).mapValues(subDirFiles => subDirFiles.map(_.file)).toList
 
     fileGroups.traverse[Hdfs, Copied] { case (subDir, sources) =>
-      pushToSubDir(sources, subDir, destRoot, archRoot, hdfsArchRoot)
+      pushToSubDir(sources.map(source => new File(source)), subDir, destRoot, archRoot, hdfsArchRoot)
```
vineethvarghese: Can this be "sources.map(new File(_))" ?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+   *
+   * @param parseH                  : function that parses the header
+   * @param parseT                  : function that parses the trailer
+   * @return (Header,Trailer)       :  parsed header and trailer values
```
SamRoberts: My comment about return types in the scaladoc comments (why do we need them? doesn't scaladoc print the return types out anyway?) apply here too, and probably other places throughout the PR.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: I don't see the point of using three parameter lists here. I see the point of creating a separate parameter list for the file, as you'll tend to be mapping over lists of files, but I think the two parsers should be combined into the same parameter list.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
jonjuscorpion666: @SamRoberts . I thought of curring the function as its a exposing function and user can easily apply within map etc

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: I should add that we probably aren't consistent with this throughout maestro, and I think as a team we've been evolving the way we write for comprehensions. But over time we seem to be settling on a style like:

```
some_stuff_here = for {
  a <- action1
  b <- startAction {
         action2
         action3
       }
  c  = expression
  d <- action4(parameter)
  e <- action5(
         bigParameter
       )
} yield e

File: maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala (hunk 11 above)
Comment (SamRoberts) [inline maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala]: Sounds good :+1:

File: maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala (hunk 8 above)
Comment (vineethvarghese) [inline maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala]: Actually @SamRoberts is right. There return type is already printed.

File: maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala [hunk 31]
```diff
@@ -75,100 +77,207 @@ case class UploadConfig(
...
+    _ <- Execution.from {
+      copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ")}
+      logger.info(s"Upload ended")
+    }
   } yield UploadInfo(copied.map(_.dest.toString))
+
+  /** Extract Header from file and Parse */
+  def hParser[H](parseH: String => Result[H], file: String): Result[H] = {
+    val inputHeader = new BufferedReader(new FileReader(file))
+    val header = Result.ok(inputHeader.readLine())
```
SamRoberts: You should be thinking about what happens if `readLine` throws an exception, and reading the `Result` documentation to see what method is best to use here. Your goal here is to ensure that, if anything in this method fails, you return a failing `Result` value, rather than throwing an exception.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
```diff
@@ -21,12 +21,11 @@ import scala.util.matching.Regex
 
 import scalaz._, Scalaz._
 
+import au.com.cba.omnia.maestro.core.task.Data
```
SamRoberts: Isn't this causing a cyclic dependency?

Now that I think of it, maybe we should be keeping `Data` in this file, and importing it into `UploadExecution.scala`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: yes, this should be grouped with the other apache commons io import

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.commons.io.FileUtils
+import au.com.cba.omnia.maestro.core.split.Splitter
+
+/* Trailer information - RowType, RecordCount, CheckSumValue, CheckSumColumn */
+case class Trailer(count: Long, checksum: Long, checksumColumn: String)
+
+/* Header information - RowType, BusinessDate, ExtractTime, FileName */
+case class Header(businessDate: String, extractTime: String, fileName: String)
+
+object ParserHT {
```
SamRoberts: if we are going to expose this as a public facing object, it will need some scaladoc, and an entry in the maestro-api project

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  /** splitting header and returning content as a Header */
+  def parseH(header: String): Result[Header] = {
+    val headerSplit = Splitter.delimited("|").run(header)
```
SamRoberts: I guess, given the lack of easy alternatives, I will have to be ok with hardcoding `Splitter.delimited("|")` for now, even though we specify the splitter to use in `LoadConfig`. It's very annoying though. Can you please raise a maestro issue  about this?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
jonjuscorpion666: We are not exposing ParseH and ParseT . Its called internal by default.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
jonjuscorpion666: Sure

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: We need to make sure that this function returns a failing `Result` rather than throwing an exception. There are a few ways you can do this.

SamRoberts: Was going to leave you to come up with a safe expression, but actually it seems like quite a nice place to show the benefits of `case` statements.

```
Splitter.delimited("|").run(header) match {
  case List("H", busDate, extractTime, fileName) => Result.ok(Header(busDate, extractTime, fileName))
  case _                                         => Result.fail(s"Failed to parse header. Expected [H|<business_date>|<extract_time>|<file_name>], but got [$header]")
}
```

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
...
+    if(headerSplit(0).trim != "H" || headerSplit.length != 4)
+      Result.fail(s"Header Parsing Failed!!!- ${header} -Please Check")
```
SamRoberts: Please don't use exclamation marks in error messages, and treat the sentences like ordinary english sentences with normal capitalization rules. Also, "please check" doesn't tell me what went wrong. I suggest an alternative error message in my code snippet above.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
...
+    else
+      Result.ok(Header(headerSplit(1), headerSplit(2), headerSplit(3)))
+  }
+
+  /** splitting trailer and returning content as a Trailer */
+  def parseT(trailer: String): Result[Trailer] = {
+    val trailerSplit = Splitter.delimited("|").run(trailer)
```
SamRoberts: my comments in `parseH` apply here too

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +34,11 @@ Upload execution properties
   can custom upload using execution monad                 $custom
   can use continue to stop after an empty upload          $continue
   can upload sequence files for the same timestamp        $sequenceFiles
+  can returns the  files matching the file pattern        $matching
```
SamRoberts: please revise these description to have correct use of singular and plural words, avoid double spaces between words, use proper english, correct capitalization, etc.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +34,11 @@ Upload execution properties
...
+  can upload the files and returns  upload info           $uploadfile
+  can Header and Trailer parsed and return right Result   $headertrailerparsing
```
SamRoberts: `Header and Trailer parsed and return right` -> `parse the header and trailer and return the right`

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +34,11 @@ Upload execution properties
...
+  can Header parsed and return right Result               $headerparsing
```
SamRoberts: my comment on the line above applies here too

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +34,11 @@ Upload execution properties
...
+  can Trailer parsed and return right Result              $trailerparsing
```
SamRoberts: Given the code we have reviewed, which will throw exceptions rather than returning `Result` failure values, it seems appropriate to test that we return the appropriate `Result` failures on garbage values, rather than throwing exceptions.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: I guess that is fine for now. If we choose to expose it in the future, there won't be any concerns around backwards incompatibility, so that sounds good. :+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +74,209 @@ case class UploadConfig(
...
+   * @return      : Returns a non-empty list of sources if we should continue the load
+   */
+  def upload(config: UploadConfig): Execution[UploadInfo] = {
+    UploadEx.matcher(config).flatMap(UploadEx.uploader(config, _))
+  }
+
+  /**
+   *  Extract the header from the file and returns header details
+   *
+   *  Header Parser function splits and parse the delimited file header string and returns trailer information
```
SamRoberts: For reference, this is how this scaladoc comment will look like on the web:

Header Parser functions splits and parse the delimited file header string and returns trailer information Default parser is implemented with assumption that Header follow pattern specified in au.com.cba.omnia.maestro.core.task.UploadExecution.parseHT User can write custom function if default function doesn't satisfy their requirment ie if file header is not in the default pattern

Are you happy with that?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +74,209 @@ case class UploadConfig(
...
+   * @return         : returns header details
+   */
+  def parseHeader[H] (parseH: String => Result[H] = ParserHT.parseH(_)) (file: String): Result[H] =
+    UploadEx.hParser(parseH, file)
+
+  /**
+   *  Extract the trailer from the file and returns trailer details
+   *
+   *  Trailer Parser function splits and parse the delimited file trailer string
+   *  and returns trailer information. Parser is implemented with assumption that
```
SamRoberts: Nice to see the proper sentence starting the paragraph here. :+1: 

Apart from that, my question about the `parseHeader` scaladoc applies here too.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -75,100 +74,209 @@ case class UploadConfig(
...
+   * Extract the header and trailer from the file and returns
+   * parsed value of header and trailer.
+   *
+   * Default parser function is designed with assumption that header and
+   * trailer matches below pattern.
+   *
+   * H|BusinessDate|ExtractTime|FileName
+   * T|RecordCount|CheckSumValue|CheckSumColumn
+   *
+   * User can write custom function if default function doesn't satisfy the requirment
```
SamRoberts: requirment -> requirement (missing an "e")

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,40 @@
...
+case class Trailer(count: Long, checksum: BigDecimal, checksumColumn: String)
+
+/* Header information - RowType, BusinessDate, ExtractTime, FileName */
+case class Header(businessDate: String, extractTime: String, fileName: String)
+
+object ParserHT {
+
+  /** splitting header and returning content as a Header */
+  def parseH(header: String): Result[Header] = {
+    Result.safe {Splitter.delimited("|").run(header)}.flatMap {
```
SamRoberts: Good idea with the `Result.safe` :smile: 

However, although I appreciate the thought behind this particular change, in this case we know we are running `Splitter.delimited`, and I happen to know that `Splitter.delimited` should not throw an exception. So I would probably not be doing `Result.safe` here.

The difference between this situation and the other one, from my point of view, is that in the other situation we were running user-supplied callback functions, but here we are running out own code, which in this case is meant to not throw exceptions.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/task/UploadExecutionSpec.scala
```diff
@@ -29,6 +34,12 @@ Upload execution properties
   can custom upload using execution monad                 $custom
   can use continue to stop after an empty upload          $continue
   can upload sequence files for the same timestamp        $sequenceFiles
+  can return the files matching the file pattern          $matching
+  can upload the files and return upload info             $uploadFile
+  can parse  header and trailer and return right result   $headerTrailerParsing
+  can parse header and return right result                $headerParsing
+  can parse trailer and return right result               $trailerParsing
+  parse garbage data and return result failure            $garbageParsing
```
SamRoberts: good :smile:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: And also, we know that the pattern matching won't fail, because we cover all the cases, and we know that simply constructing `Header` and `Trailer` values won't fail. So this function is pretty safe.

vineethvarghese: @jonjuscorpion666 Left for the day but he said that `Splitter.delimited("|").run(header)` threw a NPE when testing with an empty string.

SamRoberts: Wow, really? We should fix that

SamRoberts: I can't replicate that :frowning:

SamRoberts: It will throw a NPE when passed a null pointer, which I think is ok. But empty string seems fine.

vineethvarghese: hmm.....then let him update this with a comment.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,40 @@
...
+        case _ => Result.fail(s"Failed to parse header. Expected [H|<business_date>|<extract_time>|<file_name>], but got [$header]")
+      }
+    }
+  }
+
+  /** splitting trailer and returning content as a Trailer */
+  def parseT(trailer: String): Result[Trailer] = {
+    Result.safe {Splitter.delimited("|").run(trailer)}.flatMap {
+      _ match {
+        case List("T", count, checksum, checksumColumn) => Result.ok(Trailer(count.toLong, BigDecimal(checksum), checksumColumn))
```
rakheshrnair: A space before or after the 'checksum' value will throw NumberFormat exception when we convert it to BigDecimal. Trimming it before converting to BigDecimal would be good

SamRoberts: thanks @rakheshrnair 

Do you know why we convert checksum to BigDecimal at all? As opposed to BigInteger? Also, a checksum is often represented with hexadecimal, which I think these constructors won't support. We don't get any checksums like that?

rakheshrnair: @SamRoberts For some files we may calculate the checksum of fields whose result can give decimal values( example can be checksum of interest rates). So it will be better to keep them as BigDecimal

SamRoberts: What, you sum up the field values, or something like that? This doesn't sound like a checksum to me. I wonder if it is better to keep it as a string for now, to be safe.

SamRoberts:
Have done a fresh PR for this here: https://github.com/CommBank/maestro/pull/351

PR #351, opened 2015-04-01
Allow users to validate header and trailer information

SamRoberts:
Split upload into findSources and uploadSources, to allow users to
inspect sources before uploading them. Add parse functions for parsing
the header and trailer lines in source files.


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -14,26 +14,23 @@
...
 import scalaz._, Scalaz._
 
 import org.apache.log4j.Logger
 
-import org.apache.hadoop.conf.Configuration
+import org.apache.commons.io.input.ReversedLinesFileReader
 
 import com.twitter.scalding._
 
 import au.com.cba.omnia.omnitool.Result
```
vineethvarghese: Should there be an empty line after this import?

SamRoberts: yes, will fix

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -56,16 +54,13 @@ case class UploadInfo(files: List[String]) {
...
-  * @param controlPattern:   The regex which identifies control files. Defaults to
-  *                          detecting files starting with `S_` or files ending with
-  *                          `CTR` or `CTL` (case insensitive).
+  * @param localIngestPath  The local ingestion path where source files are found.
+  * @param hdfsLandingPath  The HDFS landing path where files are copied to.
+  * @param localArchivePath The local archive path where files are archived.
+  * @param hdfsArchivePath  The HDFS archive path, where files are also archived.
+  * @param tablename        The table name or file name in database or project.
+  * @param filePattern      The file pattern, used to identify files to be uploaded.
+  * @param controlPattern   The regex which identifies control files.
```
vineethvarghese: Don't we want the defaults mentioned?

SamRoberts: I've got to admit I was a little lazy here. The text for the default filePattern value was mucking up the rest of the scaladoc comment, so I just took it out. Can have another look.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -131,12 +140,98 @@ trait UploadExecution {
...
+    *
+    * @tparam H Header data type
+    * @tparam T Trailer data type
+    *
+    * @param parseH Function that parses the header
+    * @param parseT Function that parses the trailer
+    *
+    * @return Header and trailer data
+    */
+  def parseHT[H, T](parseH: String => Result[H] = ParserHT.parseH, parseT: String => Result[T] = ParserHT.parseT)(data: DataFile): Result[(H, T)] = {
```
vineethvarghese: Should we rename this method to parseHeaderTrailer and leave the HT convention just in ParserHT?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: fixed

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Hrmm, I am open to doing so, but I have to admit I actually thought `HeaderTrailer` was a bit too verbose. Besides, eventually we are going to be exposing `ParserHT` under some name, and do you really want to have `HeaderTrailer` in that object name too?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -131,12 +140,98 @@ trait UploadExecution {
...
+    for {
+      header   <- UploadEx.hParser(parseH, data.file)
+      trailer  <- UploadEx.tParser(parseT, data.file)
+    } yield(header, trailer)
```
vineethvarghese: I am just thinking aloud here but in this implementation if hParser and tParser has parse issues, we would first signal an issue with parseH and then signal the parseT issue when the user calls this method the second time(i.e. after fixing their header).

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +241,41 @@ trait UploadExecution {
...
+             ))
+    _     <- Execution.from(
+               files.controlFiles.foreach(ctrl => logger.info(s"skipping control file ${ctrl.file.getName}"))
+             )
+  } yield files.dataFiles
+
+  def uploader(conf: UploadConfig, files: List[DataFile]): Execution[UploadInfo] = for {
+    _    <- Execution.from(logger.info("Starting fileUploads"))
+              copied <- Execution.fromHdfs(
+              Push.push(files, conf.hdfsLandingPath, conf.localArchivePath, conf.hdfsArchivePath)
```
vineethvarghese: Indentation missing.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Good point, but I think the proper solution to this probably involves a new combinator being added to `ResultantMonad`. I think I will create an issue for that, rather than fix it now.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: thanks, will fix

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +241,41 @@ trait UploadExecution {
...
+              copied.foreach { case Copied(files, dest) => logger.info(s"Copied ${files.map(_.getName).mkString(", ")} to $dest ")}
+              logger.info(s"Upload ended")
+            }
   } yield UploadInfo(copied.map(_.dest.toString))
+
+  def hParser[H](parseH: String => Result[H], file: String): Result[H] =
+    Result.safe {
+      val reader = new BufferedReader(new FileReader(file))
+      try     { reader.readLine }
+      finally { reader.close }
```
vineethvarghese: Don't we have a `Closeable` trait somewhere? I will find out where it is.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +241,41 @@ trait UploadExecution {
...
+    }.flatMap(parseH(_))
+
+  def tParser[T](parseT: String => Result[T], file: String): Result[T] =
+    Result.safe {
+      val reader = new ReversedLinesFileReader(new File(file))
+      try     { reader.readLine }
+      finally { reader.close }
```
vineethvarghese: Same comment as above.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+  * WARNING: Apart from the behavior of parseH and parseT, not considered part
+  * of the maestro API. We may not maintain backwards incompatibility.
+  */
+object ParserHT {
+  /** Default header parser */
+  val parseH: String => Result[Header] =
+    (header: String) => Splitter.delimited("|").run(header) match {
+      case List("H", busDate, extractTime, file) => Result.safe(Header(busDate, extractTime, file))
```
vineethvarghese: Why not use Result.ok here? Do you expect anything to fail here? Result.safe will work too :)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: readers are instances of Closeable. Is there some method I can use to apply this pattern for me?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: No, it was to be symmetrical with the footer parse function, which does `toLong` on a string, and hence might fail.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
vineethvarghese: https://github.com/CommBank/omnitool/blob/master/core/src/main/scala/au/com/cba/omnia/omnitool/Closeable.scala It  does really save a lot so I will leave it up to you.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
rowandavies: That seems appropriate.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/ParserHTSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  split the header and return Header case class   $splitheader
+  split the trailer and return Trailer case class $splittrailer
+  split the header with more number of element    $splitheaderlong
+  split the trailer with more number of element   $splittrailerlong
+  split the header with wrong type format         $splitwrongheaderformat
+  split the trailer with wrong type format        $splitwrongtrailerformat
+"""
+
+  def splitheader =
```
vineethvarghese: Camel case for test names in this file?

vineethvarghese:
:+1: Happy with this once you have addressed the last few comments in there.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Yeah. If that were the standard way we have used closeables throughout our projects, I would probably do the same thing here for the sake of consistency. But I think, in isolation, I don't really agree with bringing all that machinery to bear on saving a single line of code (and making it a more complex line at that).

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/ParserHTSpec.scala
SamRoberts: Will fix

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: And same response :smile:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,37 @@
+package au.com.cba.omnia.maestro.core.upload
+
+import java.io.{FileReader, BufferedReader}
+
+import scala.math.BigDecimal
+
+import org.apache.commons.io.input.ReversedLinesFileReader
```
rowandavies: Is `ReversedLinesFileReader` actually used in the code below?

SamRoberts: Well spotted! No, it's not, that code must have originally been here and then moved to UploadExecution. Fixed.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
```diff
@@ -0,0 +1,34 @@
...
+
+/** Standard header information */
+case class Header(businessDate: String, extractTime: String, fileName: String)
+
+/** Standard trailer information */
+case class Trailer(count: Long, checkValue: String, checkValColumn: String)
+
+/**
+  * WARNING: Apart from the behavior of parseH and parseT, not considered part
+  * of the maestro API. We may not maintain backwards incompatibility.
```
rowandavies: I don't fully understand this comment.  It suggests that the behaviour of `parseH` and `parseT`is considered part of th maestro API.  But, what else does the code below define?

SamRoberts: The name of the object, the name of the methods. We might change these when we make the object public.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
```diff
@@ -146,29 +243,41 @@ trait UploadExecution {
...
+  def hParser[H](parseH: String => Result[H], file: String): Result[H] =
+    Result.safe {
+      val reader = new BufferedReader(new FileReader(file))
+      try     { reader.readLine }
+      finally { reader.close }
+    }.flatMap(parseH(_))
+
+  def tParser[T](parseT: String => Result[T], file: String): Result[T] =
+    Result.safe {
+      val reader = new ReversedLinesFileReader(new File(file))
```
rowandavies: I had the impression from discussions I've had that the trailer is not always the last line in the file - sometimes there are extra blank lines at the end, or similar.  I thought it was always the last line starting with `T`.  But, at this stage I guess I should just double check what I was told and raise an issue if it seems appropriate.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
rowandavies: Oh - okay.  It still seems strange to suggest that external clients can depend on the behaviour of `parseH` and `parseT` not changing.  If those names become undefined, I usually think of it as a change in behaviour.  Maybe just clarify quickly?  Up to you.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Ok, thanks for mentioning. Would you mind doing that?

rowandavies:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/ParserHT.scala
SamRoberts: I'll just try to come up with decent names and expose the object, probably easier to do than some strange disclaimer.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/UploadExecution.scala
SamRoberts: Fixed

PR #354, opened 2015-04-07
added aliases for fromFoo functions in RichExecution

SamRoberts:


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
```diff
@@ -61,6 +61,10 @@ case class RichExecution[A](execution: Execution[A]) {
 
 /** Pimps the Execution object. */
 case class RichExecutionObject(exec: Execution.type) {
+  /** Alias for [[fromHive]] */
+  def hdfs[T](hdfs: Hdfs[T]): Execution[T] =
```
rowandavies: Consider changing the name of the parameter to be different from the name of the function, just to avoid it being mistaken for recursion in the body - and similarly for the others below, particularly `hive`.

stephanh:
:+1:

rowandavies:
I like the shorter names.  I was worried that e.g., `hdfs(...)` might be mistaken for an application of a companion object `Hdfs(...)`, but considering where `fromHdfs(...)` is commonly used, this seems unlikely in practice.  

Just one suggestion which I've commented on in the code related to having parameters like `hdfs` with the same name as the function.

rowandavies:
:+1:

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/scalding/RichExecution.scala
SamRoberts: Good point.

PR #359, opened 2015-04-10
standardise on slf4j logging

### LINKED ISSUE #308: standardise on slf4j instead of log4j
SqoopExecution and UploadExecution use log4j, instead of slf4j, which we use elsewhere. There's no reason to have two different logging frameworks. We should just standardise on slf4j.

SamRoberts:
This fixes #308


shmookey:
Looks good to me

PR #362, opened 2015-04-13
TimeSource.predetermined does not handle timezones

### LINKED ISSUE #265: "now" time source uses UTC, unlike other time sources
The time source returned by the "now" function converts the current time to UTC. This is different to all other existing time sources I am aware of. Should it be changed to be consistent with out other time sources?

Please refer to #204 for discussion about how we should change the time zone behavior of our time sources in general. This issue is specific to the "now" function.

SamRoberts:
This fixes #265. Although this is arguably a regression,
it is in line with our other dates. This commit aims for consistency,
rather than providing a solution to our time zone issues.


rowandavies:
@SamRoberts I can't think of any places where this will cause issues - going this way, the digits in "now" will suddenly jump a bit ahead.   Representing date-times as strings this way seems unfortunate, but this change at least makes it more consistent.  :+1:

PR #364, opened 2015-04-14
allow splitters to fail

SamRoberts:


stephanh:
Considering that this is in the most called part of our code I'm concerned about the performance impact that might have.

SamRoberts:
I find it hard to believe the performance impact is material next to our decoders. I think we need to get the benchmark project up and running again if we are going to talk about these issues.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/task/LoadExecution.scala
```diff
@@ -218,21 +218,20 @@ object LoadEx {
...
-      .flatMap { case (parts, fp) => conf.filter.run(parts) match {
-        case Some(r) => Some(r)
-        case None    => {
-          // Uses Cascading counter instead of Scalding. See `AddFlowProcess` for details.
-          fp.increment("maestro", "tuples_filtered", 1L)
-          None
-        }
-      }}.map(decodeRow[A](conf))
+      .map { case (row, fp) => (conf.splitter.run(row.line).map(_ ++ row.extraFields), fp) }
+      .flatMap { case (result, fp) => result.map(conf.filter.run(_)) match {
```
stephanh: [Codacy](https://www.codacy.com/app/CommBank/maestro/file/471435981/issues/source?bid=2205397&fileBranchId=2248070#l222) Default case not specified.

SamRoberts:
The benchmarks I have run show that wrapping the splitter output in `Result` makes very little difference to the performance of the function. I don't have any benchmark results for the changes in the overall load pipeline, though.

vineethvarghese:
@SamRoberts what do you want to do with this PR?

SamRoberts:
I would like to get it merged in. I am just waiting for reviewers :smile: 

If you want major changes, I'd probably best hand it over to you guys. If you just want minor modifications, I can do them.

vineethvarghese:
@SamRoberts I will take a look at it today. Do you think we should do more performance tests on this?

SamRoberts:
@vineethvarghese I was never really of the opinion that it would have a big effect. You'd probably want to ask @stephanh to see what he thinks.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/split/Splitter.scala
```diff
@@ -45,7 +40,14 @@ object Splitter {
...
     */
   def delimited(delimiter: String) =
-    Splitter(s => GoogSplitter.on(delimiter).split(s).asScala.toList)
+    Splitter(DelimitedParseFunction(delimiter))
+
+  private case class DelimitedParseFunction(delimiter: String) extends Function[String, Result[List[String]]] {
+    // splitter will not be created until function is used, and so does not have to be serialized
+    lazy val splitter = GoogSplitter.on(delimiter)
+    def apply(line: String): Result[List[String]] =
+      Result.ok(splitter.split(line).asScala.toList)
```
vineethvarghese: Why not `Result.safe`? I assume this was because it always succeeded before?

SamRoberts: Yeah, I would be happy with `Result.safe`, but there is no particular reason to believe the delimited splitter will ever fail, whereas the csv splitter has well defined input it should fail on

SamRoberts:
Should I close this?

vineethvarghese:
Maybe leave it as it is. We want to do it but it is just not the most important thing at this time.

SamRoberts:
Ok, I'll leave it with you

vineethvarghese:
Tracking this here https://github.com/CommBank/maestro/issues/485

PR #375, opened 2015-04-23
Adding a function to partition using static value and date field

jithinjohn:
This allows to use a static value in partition along with date.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
@@ -60,6 +60,13 @@ object Partition {
       (dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d", f"${dt.getHourOfDay}%02d")
     }, "%s/%s/%s/%s")
 
+  /** Partition by static column, year, month, day for a given dateFormat. */
+  def byFieldPlusDate[A](staticField: (String, String), date: Field[A, String], dateFormat: String = "y-M-d"): Partition[A, (String, String, String, String)] =
+    Partition(List(staticField._1, "year", "month", "day"), v => {
+      val dt = DateTimeFormat.forPattern(dateFormat).parseDateTime(date.get(v))
+      (staticField._2, dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d")
+    }, "%s/%s/%s/%s")
+
   /**
     * Prepends `partition_` to column names to avoid clashes between core columns and members of the
     * thrift struct.


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
```diff
@@ -60,6 +60,13 @@ object Partition {
       (dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d", f"${dt.getHourOfDay}%02d")
     }, "%s/%s/%s/%s")
 
+  /** Partition by static column, year, month, day for a given dateFormat. */
+  def byFieldPlusDate[A](staticField: (String, String), date: Field[A, String], dateFormat: String = "y-M-d"): Partition[A, (String, String, String, String)] =
```
stephanh: Codacy Issue found: [File line length exceeds 140 characters](https://www.codacy.com/app/CommBank/maestro/file/575624492/issues/source?bid=2205397&fileBranchId=2269668#l64)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
```diff
@@ -60,6 +60,13 @@ object Partition {
       (dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d", f"${dt.getHourOfDay}%02d")
     }, "%s/%s/%s/%s")
```
stephanh: Codacy Issue found: [The string literal "%s/%s/%s/%s" appears 2 times in the file.](https://www.codacy.com/app/CommBank/maestro/file/575624492/issues/source?bid=2205397&fileBranchId=2269668#l61)

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
toddmowen: Calling this a "static column" confused me. Would it be more accurate to call it a "constant value"?

stephanh:
@jithinjohn this looks like a very specific bespoke piece of code. What is the value of adding it to Maestro? Also I don't understand why do you need to partition on a static field which is not part of the data?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
jithinjohn: Sorry for the confusion. The reason for choosing name "static"  is due to static partition feature provided by hive. The value is a constant and we can change the comment

jithinjohn:
@stephanh For couple of new sources we do have the requirement to use static partition feature of hive and need to use a static field in partition. More than one project needs to use this feature. So we thought it will be good to add this in maestro so that it could be used by similar projects.

stephanh:
@jithinjohn can you please provide the context for why you need to do that?
@anandsiva are you accross this?

As it stands the partition function you have is too bespoke to go into Maestro. There might be some more generic functionality that can help you that we can explore or we can consider putting it in etl-util. I'm just trying to get my head around what you actually need to do.

anandsiva:
@stephanh I'm not across this requirement
@jithinjohn Can you please provide details on why you need this kind of partition?

jithinjohn:
@stephanh @anandsiva we have new loads where files will be sent from different regions like syndey, newyork etc which will have same structure. We will be loading them into same hive table but need to use a static partition `region=regionName`  like `region=SY` along with data field. To support this we need to have this partition.

jithinjohn:
@stephanh Could you please confirm whether we could have this partition in maestro or should we move this to etl.util

stephanh:
etl.util

stephanh:
Closing since this code should either be more generic or go into etl-util.

Issue #384, opened 2015-05-07
Automap macro transform function restricting number of parameters to 22


sudjam:
Automap transform function throws error when more than 22 parameters are passed to it. For 24 parameters, we get the error like "object Tuple24 is not a member of package scala".

When we tried to investigate the issue, we came across http://stackoverflow.com/questions/4152223/why-are-scala-functions-limited-to-22-parameters

Please suggest if we can have work-around.

stephanh:
@sudjam are you able to use the transform macro instead in the mean time https://github.com/CommBank/maestro/blob/master/maestro-example/src/main/scala/au/com/cba/omnia/maestro/example/TransformExecution.scala#L56.

stephanh:
Scala doc: https://commbank.github.io/maestro/latest/api/index.html#au.com.cba.omnia.maestro.macros.Macros$

sudjam:
@stephanh Yes.. transform macro works. It accepts more than 22 parameters. Thanks.

shmookey:
This issue is addressed in https://github.com/CommBank/maestro/pull/388 
Automap now uses a slightly different notation: braces instead of brackets, and no commas at the end of each rule. An example is available at https://github.com/CommBank/maestro/blob/e4834b09b46746f63d1aeaa2a67cab279003af9c/maestro-example/src/main/scala/au/com/cba/omnia/maestro/example/CustomerJob.scala

PR #388, opened 2015-05-13
Support block syntax for large numbers of transform rules.

shmookey:
Addresses https://github.com/CommBank/maestro/issues/384 by adding support for block-style assignment of transform rules.

### DIFF
--- a/maestro-macros/src/main/scala/au/com/cba/omnia/maestro/macros/Automap.scala
+++ b/maestro-macros/src/main/scala/au/com/cba/omnia/maestro/macros/Automap.scala
@@ -27,14 +27,14 @@ import scala.collection.immutable.StringOps
   * If more than one such input candidate exists, all structs with a field by this name must
   * agree on its value, otherwise a runtime exception is raised.
   *
-  * To use the automapper, declare a method with the desired signature, assign to it a tuple of
-  * transformation rules and annotate it with @automapper. The macro fills in the method with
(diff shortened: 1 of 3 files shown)


maestro-macros/src/main/scala/au/com/cba/omnia/maestro/macros/Automap.scala
```diff
@@ -215,7 +216,11 @@ object automap {
         
     /** Pull apart the source object, pad it out, put it back together. */
     def mkMapper(src: Tree): Tree = {
-      val q"def $name (..$srcArgs): $srcTo = (..$srcRules)" = src
+      val q"def $name (..$srcArgs): $srcTo = $srcBody" = src
+      val srcRules = srcBody match {
```
stephanh: Codacy Issue found: [Default case not specified.](https://www.codacy.com/app/CommBank/maestro/file/634899072/issues/source?bid=2205397&fileBranchId=2309073#l220)

stephanh:
So the difference between block and tuple style is using `{}` instead of `()` and no trailing `,`?

In that case can we just use the block style?

shmookey:
@stephanh yep, done.

rowandavies:
The syntax with blocks seems nicer, because no commas are required.  I guess multi-line expressions after the `:=` will now require parens or braces, right?

shmookey:
@rowandavies yep

PR #393, opened 2015-05-19
Do not export compile-time macro dependencies. This commit fixes #355

### LINKED ISSUE #355: Macro dependencies are pulled in at runtime
Libraries which depend on maestro are going to pull in all maestro's dependencies at runtime. This includes macro dependencies which are only required at compile time.

Ideally there would be a reasonable way of importing maestro which allowed me to compile against all dependencies, including macro dependencies, but only include non-macro dependencies at runtime.

triggerNZ:


kfish:
LGTM :+1:

SamRoberts:
Do downstream projects which build against maestro get the macros on their compile time classpath, but not on their runtime path?

triggerNZ:
After further investigation, I haven't proven that this works. It was working due to the macro dependencies being pulled in transitively from other projects.

triggerNZ:
Closing, since we have to refactor macro dependencies in dependent projects too, which is scope beyond this PR

Issue #391, opened 2015-05-18
Add tests for automap error messages


stephanh:
We should test that when automap fails to compile it produces the right error messages. See the tests in JoinMacroSpec.

PR #395, opened 2015-05-21
Add tests for automap failures.

triggerNZ:
This commit fixes issue #391.

The tests compile code at compile time to check for macro expansion errors. This cannot be done using
com.twitter.util.Eval because it doesn't support macro annotations out of the box. It would be nice
to fix that, but for now this approach works.


triggerNZ:
Closing this, as #396 includes this commit and more. Will reopen if there is an issue with it

Issue #385, opened 2015-05-07
Load Execution returning EmptyLoad even if valid records available after filtering


SajeevAr:
In a file based load, the type returned in LoadInfo is used for handling different load scenarios. But LoadExecution is returning an EmptyLoad even though valid records are available after filtering. While running the same job again with same input files, it may return LoadSuccess. This issue is occurring only randomly.

It seems this is due to a TimeoutException occurring in HadoopStepStats, while attempting to fetch counter values for the job. TimeoutException is handled in HadoopStepStats.java and it only logs a warning as shown below. The warning is only found in logs where this issue has occured and it is not seen for any other valid EmptyLoad/LoadSuccess scenarios.

Please find the relevant log below :

15/05/06 14:30:58 INFO flow.Flow: [] starting
15/05/06 14:30:58 INFO flow.Flow: []  source: Hfs["TextLine[['offset', 'line']->[ALL]]"]["/dev/source/commsee/car/CSE_CCA_PRICING_OPTION/2015/05/05"]
15/05/06 14:30:58 INFO flow.Flow: []  sink: Hfs["CombinedSequenceFileScheme[[0:1]]"]["/tmp/hadoop-acoe_omnds_dintg/maestro/7e5cbe13-fdb1-47c7-8d96-eca60fad71a6.seq"]
15/05/06 14:30:58 INFO flow.Flow: []  parallel execution is enabled: true
15/05/06 14:30:58 INFO flow.Flow: []  starting jobs: 1
15/05/06 14:30:58 INFO flow.Flow: []  allocating threads: 1
15/05/06 14:30:58 INFO flow.FlowStep: [] starting step: (1/1) ...7c7-8d96-eca60fad71a6.seq
15/05/06 14:31:05 INFO mapred.FileInputFormat: Total input paths to process : 1
15/05/06 14:31:19 INFO flow.FlowStep: [] submitted hadoop job: job_201505030350_10719
15/05/06 14:31:19 INFO flow.FlowStep: [] tracking url: 
15/05/06 14:32:28 WARN hadoop.HadoopStepStats: fetching counters timed out after: 5 seconds, attempts: 1
15/05/06 14:32:28 INFO util.Hadoop18TapUtil: deleting temp path /tmp/hadoop-acoe_omnds_dintg/maestro/7e5cbe13-fdb1-47c7-8d96-eca60fad71a6.seq/_temporary
[...]

PR #402, opened 2015-05-28
Adding a timeout for fetching job counters

vineethvarghese:
The timeout value is set to 60 seconds which seems sufficiently high but I am happy to change it.


vineethvarghese:
Addresses https://github.com/CommBank/maestro/issues/385 for the time being

stephanh:
Can you bump the patch version number please.

:+1:.

stephanh:
Can you also add a comment to the PR or the issue about why this is needed, etc. for documentation and what is actually happening

rowandavies:
Agreed, add a comment so it's clear why we added this.  :+1:

vineethvarghese:
Missed the version and will add the docs too.

vineethvarghese:
Added the explanation to the issue.

PR #407, opened 2015-06-04
Adds macro thrift struct constructor.

stephanh:
Adds a function to `Inspect` that abstracts constructing scrooge and humbug thrift structs and refactored the Decode, Join and Transform macros to use.


kfish:
:+1:

PR #408, opened 2015-06-09
Fix schema compile warnings

lancelet:
The update to Scala 2.11 caused some warnings in the schema project compilation. These related to:
  - unused imports,
  - one "discarded non-Unit value" due to changing a mutable map, and
  - deprecation of the Scala parser combinators from the core library.

This PR removes those warnings.

### DIFF
--- a/maestro-schema/src/main/scala/au/com/cba/omnia/maestro/schema/taste/Sample.scala
+++ b/maestro-schema/src/main/scala/au/com/cba/omnia/maestro/schema/taste/Sample.scala
@@ -39,7 +39,7 @@ case class Sample(
   /** Pretty print the sample histogram.
    *  We print it sorted, so the most frequently occurring strings are listed
    *  first */
-  def toJsonHistogram: JsonDoc = 
+  def toJsonHistogram: JsonDoc =
     JsonMap(
       histogram
         .toList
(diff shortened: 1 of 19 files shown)


maestro-schema/src/main/scala/au/com/cba/omnia/maestro/schema/taste/Sample.scala
```diff
@@ -61,12 +61,18 @@ object Sample {
...
+      val m = smap.histogram += ((str, smap.histogram(str) + 1))
+    }
 
     // If there is still space in the map then add a new entry.
-    else if (smap.histogram.size < smap.maxSize - 1)
-      smap.histogram  += ((str, 1))
+    else if (smap.histogram.size < smap.maxSize - 1) {
+      // assigning to m suppresses a Scala warning: "discarded non-Unit value"
+      //  the map is mutable, so we don't need the new reference to it
+      val m = smap.histogram += ((str, 1))
```
stephanh: I think you could also use `append` instead of `+=` and that returns unit as well.

stephanh:
@lancelet thank you for doing that!
:+1:.

maestro-schema/src/main/scala/au/com/cba/omnia/maestro/schema/taste/Sample.scala
lancelet: Hmmm... I was hoping there would be a method like that, but I couldn't see one in the API:
http://www.scala-lang.org/api/current/#scala.collection.mutable.Map
There's `put`, but it still returns an `Option`. :frowning:

PR #410, opened 2015-06-16
Adds header and trailer parsing for files on HDFS.

### LINKED ISSUE #405: Need to do header and trailer validation on HDFS
When we switch to the upload service the files will be on HDFS so we will need to do the header and trailer validation from HDFS files.

stephanh:
This closes #405.

### DIFF
--- a/maestro-scalding/src/test/resources/trailer-parser-empty-lines/local-ingest/dataFeed/normal/mydomain/mytable_20141118.dat
+++ b/maestro-scalding/src/test/resources/trailer-parser-empty-lines/local-ingest/dataFeed/normal/mydomain/mytable_20141118.dat
@@ -5,6 +5,5 @@ D|6|ff|nfg
 D|7|ww|fff
 T|4|034|name
 
-    
-	
+
 
--- a/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/HeaderTrailer.scala
+++ b/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/HeaderTrailer.scala
@@ -1,0 +1,92 @@
+//   Copyright 2014 Commonwealth Bank of Australia
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


maestro-scalding/src/test/resources/trailer-parser-empty-lines/local-ingest/dataFeed/normal/mydomain/mytable_20141118.dat
```diff
@@ -5,6 +5,5 @@ D|6|ff|nfg
 D|7|ww|fff
 T|4|034|name
 
-
```
stephanh: @SamRoberts I noticed that you had some white spaces on the lines that were ignored by your header/trailer parsing code. Is that a requirement?

@jonjuscorpion666 @anandsiva @snehalvidyan do we have files with not just trailing empty lines but trailing lines that contain whitespace?

jithinjohn: @stephanh Files we receive from Mainframe sources have  trailers which have white spaces

stephanh: By trailers you mean the trailer itself has a whitespace, which is ok. Or there are irrelevant lines at the end of the files that contain whitespaces?

jithinjohn: yeah trailer line itself will have whitespaces.

rowandavies: @stephanh I think @SamRoberts included it because I mentioned that someone in the datahub told me that sometimes there was extra whitespace in otherwise blank lines after the header.  But, I don't recall exactly who told me, and this shouldn't be treated as a requirement based that.  I was just talking to @jonjuscorpion666, he said having such whitespace isn't normal, but it does happen, and there are many other similar rare deviations that we don't handle.

stephanh: I spoke to @anandsiva and he said we shouldn't have to handle those cases. Instead we should get upstream to fix it up which I'm strongly in favour off.

vineethvarghese:
Looks fine :+1:

PR #414, opened 2015-06-18
Shadow api 412

kfish:
This is an untested set of patches that add the API calls suggested during today's maestro hackfest, mostly designed by @maxmithun


rowandavies:
@kfish @maxmithun What's the current status of this PR?  I quickly reviewed it, it seems pretty close,  shall I add comments?   The main thing is probably having tests for the logging versions in `maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/LoadExecutionSpec.scala` and the like.

maxmithun:
@rowandavies Stephan as suggested to move the same to util.etl by creating a subproject having dependency on the controller. once i create, will be moving the methods to the sub project.

vineethvarghese:
An important thing to consider is that the maestro project is public while the etl-controller is not.

vineethvarghese:
@maxmithun :+1:

PR #417, opened 2015-06-23
Ability to lift fields to a larger enclosing structure

triggerNZ:
I am not sure if the correct word for this is *lift*

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/data/Field.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/data/Field.scala
@@ -52,5 +52,5 @@ case class Field[A : Manifest, B : Manifest](name: String, get: A => B) {
    * @tparam C the type of the bigger structure
    * @return Field[C, B]
    */
-  def lift[C : Manifest](f: C => A):Field[C, B] = Field(name, f andThen get)
+  def zoom[C : Manifest](f: C => A):Field[C, B] = Field(name, f andThen get)
 }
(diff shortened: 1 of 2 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/data/Field.scala
```diff
@@ -45,4 +45,12 @@ case class Field[A : Manifest, B : Manifest](name: String, get: A => B) {
   }
 
   override def hashCode: Int = name.hashCode * 41 + structType.hashCode
+
+  /**
+   * Creates a field accessor for a higher structure, if this field is contained inside a bigger structure
+   * @param f function from the bigger structure to `B`
```
rowandavies: The type says `f` maps to `A` but the comment says 

```
   f function from the bigger structure to `B`
```

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/data/Field.scala
```diff
@@ -45,4 +45,12 @@ case class Field[A : Manifest, B : Manifest](name: String, get: A => B) {
...
+   * @tparam C the type of the bigger structure
+   * @return Field[C,B]
```
rowandavies: format: we generally put spaces after `,`s

rowandavies:
Looks like a cool addition!

Instead of `lift` you could consider `nest` or `deep` or `embed` to indicate that you're creating field accessors with that access within an nested/deep/embedded structure.

Or `expose` or similar to indicate that you're bringing part of the nested level up.

Up to you, `lift` is okay too.  Rebase and push to master when you're ready.  :+1:

rowandavies:
Or, you could call it `dive` as in dive down to the depths of the `C`.  Or `tele` meaning "at a distance".  Or `transparent` or `glass` because you can see below the surface, or `portal` since you can only see below in some places.  Or `zoom` because it makes depths seem close.  Or, just `lift` may be more familiar for such things.

PR #421, opened 2015-07-01
Ebenezer version bump

jithinjohn:
This allows ETL projects to use modified `setupHiveTestTable` function so that user can pass delimiter while creating a hive test table.


vineethvarghese:
:+1:

PR #394, opened 2015-05-19
Get test-cluster script running again with current maestro API.

dkristian:
Include example hive-site.xml in assembled jar for hive queries to run out of the box.

Fixes #392

PR #423, opened 2015-07-07
Update examples

jameshales:
There is now a test script for each of the examples, along with documentation for running them. As opposed to the unit tests, the test scripts run the examples against actual Hadoop/Hive/MySQL.

This is based on <https://github.com/CommBank/maestro/pull/394>.

Tested on a CDH5.3.0 virtual machine set up with Vagrant, with some configuration afterwards. Specifically, you need to delete a problematic JODA time JAR:
```
sudo mv /usr/lib/hadoop-mapreduce/joda-time-1.6.jar /usr/lib/hadoop-mapreduce/joda-time-1.6.jar.bak
```
and you need to add the following properties to `/etc/hadoop/conf/mapred-site.xml`:
```
<property>
    <name>hive.metastore.uris</name>
    <value>thrift://localhost:9083</value>
  </property>
  <property>
    <name>yarn.resourcemanager.address</name>
    <value>localhost</value>
  </property>
  <property>
    <name>hive.exec.dynamic.partition.mode</name>
    <value>nonstrict</value>
  </property>
```
Other than that, follow the instructions in `maestro-example/Readme.md`.


maestro-example/src/main/scala/au/com/cba/omnia/maestro/example/CustomerSqoopImportExecution.scala
```diff
@@ -49,12 +49,28 @@ case class CustomerImportConfig(config: Config) {
...
+    // Fail if there was a problem with the load
     loadSuccess        <- loadInfo.withSuccess
+    // Write out the Customers to a Hive table
     hiveCount          <- viewHive(conf.catTable, pipe)
   } yield (CustomerImportStatus(sqoopCount, loadSuccess.written, hiveCount))
 }
 
 case class CustomerImportStatus(sqoopCount: Long, loadCount: Long, hiveCount: Long)
+
+object CustomerSqoopImportExecutionJob extends MaestroJob {
```
stephanh: We could probably just turn the Execution object into a `MaestroJob` without the need of this second wrapper object?

Are you happy to do that or do you want me to do that?

stephanh:
@jameshales very nice. Thank you for that!

vineethvarghese:
@jameshales can please merge this when you get a chance

stephanh:
Incorporated as part of #435.

PR #427, opened 2015-07-16
Fixes `HiveTable.writeExecution` bugs.

### LINKED ISSUE #426: ViewExecution.viewHive fails to overwrite partitions
The following issues were faced while overwriting partitions using viewHive (with append mode as false). eg. viewHive(conf.transformationTable, pipe, false). These issues are related to the fix for CommBank/maestro#382. etl-util version used is 1.14.0-20150625002151-e89a7bc. 

1. If the table location doesn't exist, viewHive will fail with a ResultException as shown below. This occurs as Line:101 in HiveTable.scala tries to list all the files inside a directory that does not exist. But unit test is not getting failed in this case (which is not expected), the failure is seen only while running the jar. But while trying to explicitly list the same directory (outside viewHive), test fails as expected (Line:123 in CarBaseExecution.scala). The code is pushed into stash repo etl.commsee-car (branch: car_phase2).
 
2. MSCK REPAIR TABLE (Line:122 in HiveTable.scala) is trying to repair the table in default database instead of actual database. This causes NoSuchObjectException (exception shown below) while running the jar, but the job finishes with status 0.

3. As per the current approach, it seems MSCK REPAIR TABLE is not needed, as no partitions are added or deleted while overwriting. Only the files within a partition are removed and added, which does not require MSCK REPAIR TABLE query. In the above case 2, the partitions were modified as expected even though this query has failed.

4. The current approach could fail if there is no data to overwrite in an existing partition. In that case, no parquet files will be created, but the existing files in the partition needs to be removed, which may not happen as per the current approach.

stephanh:
It removes `MSCK REPAIR TABLE` statement since it is not needed. The
current use of it did not set the database and was broken.

It also changes the overwrite parts of the code to handle the case where
a table exists in the metastore but the underlying directory has been
deleted.

This fixes #426.

### DIFF
--- a/maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
+++ b/maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
@@ -16,7 +16,11 @@ package au.com.cba.omnia.maestro.task
 
 import scalaz.Scalaz._
 
+import com.twitter.scalding.Execution
+
 import org.specs2.matcher.Matcher
+
+import au.com.cba.omnia.permafrost.hdfs.Hdfs
 
 import au.com.cba.omnia.thermometer.core.{Thermometer, ThermometerSource}, Thermometer._
 import au.com.cba.omnia.thermometer.fact.PathFactoid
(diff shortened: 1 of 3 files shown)


maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
```diff
@@ -248,6 +288,26 @@ View execution properties
...
+    } yield (a, b)
+
+    executesSuccessfully(exec) must_== ((4, 4))
+    facts(
+      hiveWarehouse </> "unpart.db"  </> "unpart_table" </> "part-*.parquet" ==> matchesFile,
+      hiveWarehouse </> "unpart2.db" </> "unpart_table" </> "part-*.parquet" ==> matchesFile
+    )
+  }
+
+
```
rowandavies: format: extra blank line?

rowandavies:
Looks good to me.  :+1:

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
stephanh: That was done on purpose to separate the different sections of the files.

PR #432, opened 2015-07-23
Fixes more partitioned overwrite bugs.

### LINKED ISSUE #382: ViewExecution.viewHive doesn't overwrite the partition as expected.
the call to 'viewHive(hiveTbale, inputPipe, append = false)' doesn’t fully overwrite the partition, instead it overwrites the files in partitions which are of the same name.Due to this some files remain in the partition and result in duplicate/unwanted  records. Currently all GDW loads using this functionality are affected by this issue.

stephanh:
This fixes #382 again. This now changes partitioned files to have unique
names.

Previously, the file name were not unique which meant that sometimes old
and new filenames were the same and we would either not recognise that a
partition had been overwritten or delete a new file from a partition
because we thought it was an old file.


stephanh:
@SajeevAr  can you please have a look at this. I tried to come up with any other test scenarios that I haven't thought of. I improved the tests to cover the scenarios you mentioned.

I also adopted the fix you suggested. Great work!

stephanh:
Builds are failing due to Travis not being able to download sbt. The url returns a 404.

vineethvarghese:
https://github.com/http4s/http4s/pull/363/files is temporary fix that can be done. Although, I would imagine Travis can quickly solve this on their end.

SajeevAr:
Thanks @stephanh . This should fix the issue.

stephanh:
Are there any other problems you can think of that we should test for?

SajeevAr:
I think the test now covers most scenarios pretty well.

PR #435, opened 2015-07-28
Update examples

stephanh:
This tidies up some of the examples and also picks up @dkristian and @jameshales since I can't edit James' branch directly.

In particular it rearranges the examples as follows:
* Removes `TransformExecution` since it is now superseded by the `sutomap`
  example in `CustomerJob`.
* Renames `CustomerJob` to `CustomerAutomapJob`.
* Turns `CustomerExecution` into a job and renames it to
  `CustomerHiveJob`.
* Turns `CustomerSqoopImportExecution` into a job and renames it to
  `CustomerSqoopImportJob`.
* Updates the Readme.

This supersedes #423.


shmookey:
Looks good to me. I seem to recall @rowandavies having some ideas about keeping some of the examples as `Execution`s instead of changing them to `Job`s, might be worth asking him about it. :+1:

rowandavies:
@stephanh @shmookey I mostly thought that CustomerExecution might stay as an Execution or be removed since CustomerJob was originally a simplified version of it, extending MaestroJob, to appear in the README.  Including the automap there seems nice, as does making the original into CustomerHiveJob.  

I guess also having an example of an execution that returns something other than a `JobStatus` would be good, it seems like many have assumed that all Executions fail via the JobStatus (this seems to have improved though).

PR #442, opened 2015-09-22
Humbug benchmark

stephanh:
Changes the benchmarks to use Humbug instead of Scrooge and adds a benchmark for a 500 fields thrift struct.


maestro-benchmark/src/test/scala/au/com/cba/omnia/maestro/benchmark/thrift/Generators.scala
```diff
@@ -19,28 +19,592 @@ import org.scalameter.api.Gen
...
+      x.str491  = s"r$row-f490"
+      x.str492  = s"r$row-f491"
+      x.dbl493  = row * 500 + 492.5
+      x.int494  = row * 500 + 493
+      x.long495 = Int.MaxValue + row * 500 + 494
+      x.str496  = s"r$row-f495"
+      x.str497  = s"r$row-f496"
+      x.dbl498  = row * 500 + 497.5
+      x.int499  = row * 500 + 498
+      x.long500 = Int.MaxValue + row * 500 + 499
```
rowandavies: It's probably overkill, but if we want to test other sizes, I guess we could have a macro generating code like this.

rowandavies:
Looks good to me, just one suggestion to use macros which is probably overkill.  :+1:

maestro-benchmark/src/test/scala/au/com/cba/omnia/maestro/benchmark/thrift/Generators.scala
stephanh: Agreeded. If we need to revisit this again we should look at that.

PR #446, opened 2015-10-14
Removed Environment Variables Command Line Argument Dependency.

JosephShobin:
Environment configuration dependency on command line argument removed from maestroconfig.


rowandavies:
@JosephShobin Okay - after we discussed and looked at other options today, I guess that we can close this PR.

Whether we might want to force configuration in code with a trait for the interface to maestro seems like a separate question.  Since it would be a breaking change, we may revisit this at some point when we have other similar changes.

PR #447, opened 2015-10-15
Removes unneeded dependency on scala-compiler

toddmowen:
Macros only require scala-reflect. Avoid listing scald-compiler as a dependency, since this will contribute an unnecessary transitive dependency to downstream projects.


stephanh:
Can you bump the version number as well please.

:+1:

PR #452, opened 2015-11-09
Avoid race condition when creating unique names in parallel

dkristian:
Use scalding's [`UniqueID`](https://github.com/twitter/scalding/blob/0.13.1/scalding-core/src/main/scala/com/twitter/scalding/Stats.scala#L92) (which incorporates an atomic counter with the system time in millis) for more guaranteed unique part file names to avoid race condition that can arise when zipping executions that append to the same location.


stephanh:
@dkristian thank you for the fix.

Just using time was pretty poor. I can't believe I did that :frowning:

:+1:.

dkristian:
@stephanh does anything need to be done about the bogus codacy warning?

stephanh:
@dkristian just ignore it.

PR #454, opened 2015-11-16
Updating parlour version and test case for sqoop import

jithinjohn:
Updating parlour version to support replace/removal of new line and control-A character from string columns during Sqoop import


maestro-example/src/test/scala/au/com/cba/omnia/maestro/example/CustomerSqoopImportJobSpec.scala
```diff
@@ -76,10 +76,10 @@ Customer Sqoop Import Job test
     Class.forName("org.hsqldb.jdbcDriver")
 
     val data = List(
```
rowandavies: Perhaps just add a comment above this to say that the `hiveDropImportDelims` option removes the `\n` and `\u0001`.

rowandavies:
Looks good, just one suggestion to add a comment for the change to the test. :+1:

jithinjohn:
@rowandavies Thanks for reviewing

PR #455, opened 2015-11-18
Update the version of Thermometer to the latest

maxmithun:
Update the version of the Thermomter to the latest. Which was not updated for a while.


stephanh:
You also added some extra files?

PR #457, opened 2015-11-23
Removing cascading-hive dependency

vineethvarghese:


maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -150,21 +148,21 @@ case class UnpartitionedHiveTable[A <: ThriftStruct : Manifest](
       def moveFiles(src: Path, dst: Path): Hdfs[Unit] = for {
         _     <- Hdfs.mkdirs(dst) // Create the folder if it doesn't already exist for some reason.
         files <- Hdfs.files(src, "*.parquet")
-        uniq  =  UniqueID.getRandom.get
-        _     <- Hdfs.mkdirs(dst)
```
stephanh: Is the removal of this line intentional?

version.sbt
```diff
@@ -12,7 +12,7 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-version in ThisBuild := "2.14.4"
+version in ThisBuild := "2.14.5"
```
stephanh: Probably should bump a minor version for this, since we transitively expose some pretty big changes  and the return type of `HiveTable.source` is different as well.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethvarghese: Yes. Do you know why we have to call that twice?

version.sbt
vineethvarghese: Sure

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
stephanh: Nope. Didn't see the original call.

stephanh:
:+1:

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethvarghese: Ok. There were 2 calls there and I was wondering what was going on.

PR #458, opened 2015-11-25
Explicitly deleting the _temporary folder in a table location which i…

vineethvarghese:
…s created by cascading


stephanh:
:+1:

PR #467, opened 2015-12-08
Reverting deletions of _temporary folder

vineethvarghese:
And ignoring hidden files at the root


rowandavies:
Looks good.  :+1:

PR #474, opened 2016-01-08
[Discussion] Run job with modified args

Fantoccini:
Hi team, 

As we have handy `OminiaEnv` defined, and both [etl.util](https://github.com/CommBank/util.etl/blob/master/core/src/main/scala/au/com/cba/omnia/etl/util/SimpleMaestroJob.scala#L65) and coppersmith (https://github.com/CommBank/coppersmith/pull/175) should use it. I'm wondering:
1. If it's appropriate add `withSubArgs` function here.
2. Should we move some common things(e.g. `EtlConfiguration`, decode hive text table) to another internal library? Then both etl.util and coppersmith(or dataproducts.common) can benefit from it.

Thanks.

### DIFF
--- a/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/scalding/RichExecution.scala
+++ b/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/scalding/RichExecution.scala
@@ -19,7 +19,7 @@ import scala.concurrent.Future
 import scalaz.{\/, Monad}
 import scalaz.\&/.{These, This, That, Both}
 
-import com.twitter.scalding.{Config, Execution}
+import com.twitter.scalding.{Args, Config, Execution}
 
 import org.apache.hadoop.hive.conf.HiveConf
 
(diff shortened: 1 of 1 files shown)


stephanh:
@Fantoccini I have no problems with adding the function in the PR. Can you please update the commit title/message to reflect the actual change you are making.

The only think I want to highlight from a designing the program perspective is that we where it is in our control we don't really want to be modifying the args or the config within an execution.
* We should be keeping our configuration, etc. in something more descriptive, robust and typesafe than the config/args.
* Ideally you also want to tweak the config, args. at the top level/entry point of your program and then run your execution using that config instead of making the changes as part of the execution.

Therefore, `withSubConfig` should only ever be used where you have to modify one of the standard hadoop settings as required by a 3rd party library and that change needs to have a limited scope. In general, you want to avoid the use of `withSubConfig` or your corresponding `withArgs` wherever possible.

toddmowen:
@Fantoccini my apologies, it wasn't my intention to encourage you to submit a PR to maestro _right now_. Better to start with the first approach suggested in the [coppersmith comment](https://github.com/CommBank/coppersmith/pull/175#issuecomment-169843141), which doesn't require modification to maestro, and then once you have two or three jobs using similar code then look for the best way to refactor.

I think I understand the problem you are trying to solve here, and there are several ways to solve it, which I suggest we talk about offline.

Fantoccini:
Thanks all, move the discussion to internal project https://github.com/CommBank/DataProducts/issues/68

PR #476, opened 2016-01-15
Use beeswax for Hive monad

shmookey:


vineethvarghese:
version updated is needed otherwise :+1:

stephanh:
So now we have got 2 hive monads? One in beeswax and one in ebenezer? And nothing even ensures that we only use the beeswax ones now?

shmookey:
@vineethvarghese Good point, done. 
@stephanh Correct (see gitter channel)

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
```diff
@@ -29,7 +29,7 @@ import au.com.cba.omnia.thermometer.fact.PathFactoids._
 import au.com.cba.omnia.thermometer.hive.ThermometerHiveSpec
 
 import au.com.cba.omnia.ebenezer.ParquetLogging
-import au.com.cba.omnia.ebenezer.scrooge.hive._
+import au.com.cba.omnia.beeswax._
```
stephanh: Can we make this a qualified import? Is there lots of stuff under Beeswax that we are importing?

version.sbt
```diff
@@ -12,7 +12,7 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-version in ThisBuild := "2.17.0"
+version in ThisBuild := "2.17.1"
```
stephanh: Lets make it a minor version bump.

stephanh:
:+1:

PR #477, opened 2016-01-18
Automatic version update

toddmowen:
If the build succeeds, then I will merge this PR immediately. (This is an experiment in using wrangler to update all internal dependencies to the latest versions).

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -34,13 +34,13 @@ import au.com.cba.omnia.humbug.HumbugSBT._
 object build extends Build {
   type Sett = Def.Setting[_]
 
-  val thermometerVersion = "1.4.0-20151222033039-8cac3ec"
-  val ebenezerVersion    = "0.21.2-20151222044909-fa96b6c"
-  val omnitoolVersion    = "1.12.0-20151021050758-700b9d0"
-  val permafrostVersion  = "0.12.0-20151122230226-4082849"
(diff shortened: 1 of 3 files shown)


project/build.scala
```diff
@@ -34,13 +34,13 @@ import au.com.cba.omnia.humbug.HumbugSBT._
 object build extends Build {
   type Sett = Def.Setting[_]
 
-  val thermometerVersion = "1.4.0-20151222033039-8cac3ec"
-  val ebenezerVersion    = "0.21.2-20151222044909-fa96b6c"
-  val omnitoolVersion    = "1.12.0-20151021050758-700b9d0"
+  val thermometerVersion = "1.4.1-20160117233509-42defec"
+  val ebenezerVersion    = "0.21.3-20160118050516-0c6842d"
+  val omnitoolVersion    = "1.12.1-20160118033143-a5ca392"
   val permafrostVersion  = "0.12.0-20151122230226-4082849"
```
stephanh: Permafrost hasn't changed?

stephanh:
Very nice work @toddmowen, propagating the changes through the stack! :clap:

project/build.scala
toddmowen: Oops, missing from the updater config.

toddmowen:
@stephanh Thanks. It's just an experiment to test the waters, and even with wrangler it has been a bit laborious. As I said to @shmookey, it could be further automated, but that would be non-trivial.

PR #480, opened 2016-02-29
Omde 272 fix field type

knservis:
It looks like it would be far more useful to get the actual manifest rather than it's class. This is a problem for us as we need to evaluate the manifest of a field in our application.


vineethvarghese:
You don't have a version update.

vineethvarghese:
Please squash commits and add update the version before merging :+1:

vineethvarghese:
Please squash commits and update the version before merging :+1:

knservis:
Awesome. Can you please tell me where to change the version.

vineethvarghese:
In case you are not aware, please use fast forward merge.

stephanh:
@knservis next time please don't introduce merge commits. As you can see from our past history we use fast forward merges to create a linear git history.

Issue #481, opened 2016-02-29
Ability to load files with  file-timestamp in different timezones.


AnnTomie:
*Background:*
SAP generates filenames with UTC as default timezone. During daylight saving, files
 were generated with timestamp ```2015-10-04-02%```
 which was invalid in AEST. While setting this timestamp to joda DateTime object, upload step failed as AEST is considered as the default timezone. From the execution we use a regex pattern to identify files from inbound location and if regex matches, it extracts the timestamp from the filename. When joda accepts this timestamp in default timezone, we end up in failure scenarios during daylight savings. Next
 daylight saving date is  Apr 3, 2016 .
*Related issue* : https://github.com/CommBank/datahub/issues/132

**Feature:**
Ability to indicate the timezone associated with the filename whenever it is other than AEST format. Currently, Maestro Load (including Upload,viewHive steps ) expects file timestamp in AEST format by default.

rowandavies:
This is also related to issue #204

rowandavies:
Resolved by #483

SajeevAr:
The current timezone addition provides the ability to specify timezone associated with a file during upload and hence fixes the issues in upload step. However, apart from upload, the default AET timezone was assumed in the following instances as well, which may lead to job failures.

1. Methods `byYear`, `byMonth`, `byDay`, `byHour`, `byDate` etc. in `Partition.scala` will parse the timestamp in partition field based on a given format. So viewHive step will fail if the timestamp is invalid in default timezone.
2. `isDate` method in `Validator.scala` will fail if timestamp in validator field is invalid in default timezone.
3. `now` method in `TimeSource.scala` also use the default timezone to generate `Timesource`. It doesn't cause a job failure, but it restricts from creating timesource in a particular timezone.

To fix these issues, we need to add the capability to specify the required timezone for these scenarios as well.

andynygard:
@SajeevAr Could you please open a new issue so that we have visibility of this?

PR #483, opened 2016-05-03
Completed Time Zone additions, return `DateTime`s, use Result, bug fixes.

rowandavies:
Addresses #481

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
@@ -19,42 +19,63 @@ import java.io.File
 
 import scala.util.matching.Regex
 
+import org.joda.time.{DateTime, DateTimeZone, Period}
+
 import scalaz._, Scalaz._
+
+import au.com.cba.omnia.omnitool.Result
 
 /** A control file, not to be loaded into HDFS */
 case class ControlFile(file: File)
 
-/** A file to be uploaded, with associated business date directory*/
+/** A control file, not to be loaded into HDFS, with the timestamp and inferred frequency between timestamps */
+case class ControlFileTimestamped(file: File, timestamp: DateTime, frequency: Period) {
+  def toControlFile = ControlFile(file)
+}
+
+/** A file to be uploaded, with associated business date directory */
 case class DataFile(file: String, parsedDate: String)
 
-/** Files found by [[Input.findFiles]] */
+/** A file to be uploaded, with its business date directory, timestamp and inferred frequency between timestamps */
+case class DataFileTimestamped(file: String, parsedDate: String, timestamp: DateTime, frequency: Period) {
+  def toDataFile = DataFile(file, parsedDate)
+}
+
+/** Files found by [[Input.findFiles]] without timestamps and inferred frequencies */
 case class InputFiles(controlFiles: List[ControlFile], dataFiles: List[DataFile])
+
+/** Files found by [[Input.findFiles]] with timestamps and inferred frequencies */
+case class InputFilesTimestamped(controlFiles: List[ControlFileTimestamped], dataFiles: List[DataFileTimestamped]) {
+  def toInputFiles = InputFiles(controlFiles.map(_.toControlFile), dataFiles.map(_.toDataFile))
+}
 
 /** Factory for `Input`s */
 object Input {
 
   /**
-    * Find files from a directory matching a given file pattern.
+    * Find files from a directory matching a given file pattern, also returning `DateTime`s and frequencies.
     *
-    * See the description of the file pattern at [[au.cba.com.omnia.meastro.core.task.Upload]]
+    * See the description of the file pattern at [[au.com.cba.omnia.maestro.core.task.Upload]]
     */
   def findFiles(
-    sourceDir: String, tablename: String, filePattern: String, controlPattern: Regex
-  ): String \/ InputFiles = for {
-    matcher <- InputParsers.forPattern(tablename, filePattern)
+    sourceDir: String, tablename: String, filePattern: String, controlPattern: Regex, timeZone: DateTimeZone
+  ): Result[InputFilesTimestamped] = for {
+    matcher <- InputParsers.forPattern(timeZone)(tablename, filePattern)
     files   <- fromNullable((new File(sourceDir)).listFiles, s"${sourceDir} does not exist").map(_.toList)
     results <- files traverse (file => matcher(file.getName))
   } yield {
-    val matches      = (files zip results) collect { case (file, Match(dirs)) => (file, dirs) }
-    val (ctrl, data) = matches partition { case (file, _) => isControl(file, controlPattern) }
-    val controls     = ctrl map { case (file, _)    => ControlFile(file) }
-    val dataFiles    = data map { case (file, dirs) => DataFile(file.toString, dirs mkString File.separator) }
-    InputFiles(controls, dataFiles)
+    val matches      = (files zip results) collect { case (file, Match(dirs, stamp, freq)) => (file, dirs, stamp, freq) }
+    val (ctrl, data) = matches partition { case (file, _, _, _) => isControl(file, controlPattern) }
+    val controls     = ctrl map { case (file, _, stamp, freq)   => new ControlFileTimestamped(file, stamp, freq) }
+    val dataFiles    = data map { case (file, dirs, stamp, freq)   =>
+      new DataFileTimestamped(file.toString, dirs mkString File.separator, stamp, freq)
+    }
+    new InputFilesTimestamped(controls, dataFiles)
   }
 
   def isControl(file: File, controlPattern: Regex) =
     controlPattern.unapplySeq(file.getName).isDefined
 
   def fromNullable[A](v: A, msg: String) =
-    if (v == null) msg.left else v.right
+    if (v == null) Result.fail(msg) else Result.ok(v)
 }
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
@@ -12,20 +12,20 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-package au.com.cba.omnia.maestro.core
-package upload
+package au.com.cba.omnia.maestro.core.upload
 
 import scala.util.control.NonFatal
 
 import scala.util.parsing.combinator.Parsers
 import scala.util.parsing.input.CharSequenceReader
 
-import org.joda.time.{DateTime, DateTimeFieldType, MutableDateTime}
+import org.joda.time.{DateTime, DateTimeFieldType, MutableDateTime, DateTimeZone, Period}
 import org.joda.time.format.DateTimeFormat
 
 import scalaz._, Scalaz._
 
 import au.com.cba.omnia.omnitool.time.DateFormatInfo
+import au.com.cba.omnia.omnitool.Result
 
 /** The results of a match against a file name */
 sealed trait MatchResult
@@ -37,9 +37,9 @@ case object NoMatch extends MatchResult
   * The file matches the pattern.
   *
   * Also contains the directories for the file
-  * corresponding to the date time fields in the pattern.
+  * corresponding to the date time fields in the pattern, the date time, and the inferred frequency.
   */
-case class Match(dirs: List[String]) extends MatchResult
+case class Match(dirs: List[String], stamp: DateTime, frequency: Period) extends MatchResult
 
 /**
   * Contains parsers for input file patterns and input files
@@ -50,53 +50,36 @@ case class Match(dirs: List[String]) extends MatchResult
   */
 object InputParsers extends Parsers {
 
-  /** Either a value or an error message */
-  type MayFail[A] = String \/ A
-
-  /** Factory functions for MayFail */
-  object MayFail {
-
-    /** convert an option to a disjunction */
-    def fromOption[A](option: Option[A], msg: String): MayFail[A] =
-      option match {
-        case Some(a) => a.right
-        case None    => msg.left
-      }
-
-    /** Fail if condition is false */
-    def guard(cond: Boolean, msg: String): MayFail[Unit] =
-      if (cond) ().right else msg.left
-
-    /** Safely run an operation that may throw an exception */
-    def safe[A](a: => A): MayFail[A] =
-      try { a.right } catch { case NonFatal(ex) => ex.toString.left }
-
+  /** convert an option to a Result - expected to be incorporated into omnitool. */
+  def fromOption[A](option: Option[A], msg: String): Result[A] =
+    option match {
+      case Some(a) => Result.ok(a)
+      case None    => Result.fail(msg)
   }
 
   /**
-    * Pass in table name and file pattern and get back a matching function on file names.
+    * Pass in timezone, table name and file pattern and get back a matching function on file names.
     *
     * Both the pattern parsing and the matching functions can fail with an error message.
     */
-  def forPattern(tableName: String, pattern: String): MayFail[String => MayFail[MatchResult]] =
-    PatternParser.pattern(tableName)(new CharSequenceReader(pattern)) match {
-      case NoSuccess(msg, _) => {
-        msg.left
-      }
+  def forPattern(tz: DateTimeZone)(tableName: String, pattern: String): Result[String => Result[MatchResult]] =
+    PatternParser.pattern(tableName, tz)(new CharSequenceReader(pattern)) match {
+      case NoSuccess(msg, _) => Result.fail(msg)
       case Success(fileNameParser, _) => {
-        val func: String => MayFail[MatchResult] =
+        val func: String => Result[MatchResult] =
           (fileName: String) => fileNameParser(new CharSequenceReader(fileName)) match {
-            case Error(msg, _)    => msg.left
-            case Failure(_, _)    => NoMatch.right
-            case Success(dirs, _) => Match(dirs).right
+            case Error(msg, _)    => Result.fail(msg)
+            case Failure(_, _)    => Result.ok(NoMatch)
+            case Success((dirs, stamp, frequency), _) => Result.ok(Match(dirs, stamp, frequency))
           }
-        func.right
+        Result.ok(func)
       }
     }
 
-  /** Throws parse error (not parse failure) if disjunction failed */
-  def mandatory[A](x : MayFail[A]): Parser[A] =
-    x.fold(msg => err(msg), a => success(a))
+
+  /** Throws parse error (not parse failure) if Result failed */
+  def toParser[A](x : Result[A]): Parser[A] =
+    x.foldMessage(a => success(a), msg => err(msg))
 
   type Elem = Char
 
@@ -110,10 +93,13 @@ object InputParsers extends Parsers {
   /**
     * Input file name parser
     *
-    * Parses a file name and returns the directories which the file should be
-    * placed in.
+    * Parses a file name and returns a list of the path parts to the destination
+    * directory, (e.g., for daily files: YYYY/MM/DD) along with the
+    * corresponding `DateTime` and inferred frequency (e.g., 1 day or 1 hour).
+    * Unused components in the `DateTime` default to the least valid value
+    * (E.g., for yearly files: January 1st 00:00).
     */
-  type FileNameParser = Parser[List[String]]
+  type FileNameParser = Parser[(List[String], DateTime, Period)]
 
   /**
     * Building block of an input file name parser.
@@ -130,14 +116,14 @@ object InputParsers extends Parsers {
     /**
       * Sequential composition with another partial file name parser
       *
-      * It is a parse error (not failure) if the date time fields are consistent.
+      * It is a parse error (not failure) if the date time fields are inconsistent.
       */
     def followedBy(other: PartialParser) = {
       val combinedFields = (fields ++ other.fields).distinct
       val combinedParser = for {
         dt1 <- parser
         dt2 <- other.parser
-        dt3 <- mandatory(combineDateTimes(dt1, fields, dt2, other.fields))
+        dt3 <- toParser(combineDateTimes(dt1, fields, dt2, other.fields))
       } yield dt3
       PartialParser(combinedFields, combinedParser)
     }
@@ -147,24 +133,27 @@ object InputParsers extends Parsers {
       *
       * Given two date times, and a list of fields defined in each date time,
       * return a new date time containing the union of all fields. It is an
-      * error if the date times have different values for any field
+      * error if the date times have different values for any field, since
+      * parsers should always yield only a single date time, even if some
+      * or all components are repeated.
       */
-    def combineDateTimes(dt1: DateTime, fs1: List[DateTimeFieldType],
-      dt2: DateTime, fs2: List[DateTimeFieldType]): MayFail[DateTime] =
+    def combineDateTimes(
+      dt1: DateTime, fs1: List[DateTimeFieldType], dt2: DateTime, fs2: List[DateTimeFieldType]
+    ): Result[DateTime] =
 
       // if there are no fields in fs1, just return unmodified dt2
-      if (fs1.isEmpty) dt2.right
+      if (fs1.isEmpty) Result.ok(dt2)
 
       // if there are fields in fs1, copy fields from fs2 to fs1
       // (if there are no fields in fs2 this will quickly return unmodifed dt1)
-      else fs2.foldLeftM[MayFail, DateTime](dt1)((dtAcc, f) => {
+      else fs2.foldLeftM[Result, DateTime](dt1)((dtAcc, f) => {
         val field  = dt2.property(f).getField
         val dt2Val = field.get(dt2.getMillis)
         val dt1Val = field.get(dt1.getMillis)
         val inFs1  = fs1.contains(f)
-        if      (!inFs1)                    dtAcc.withField(f, dt2Val).right
-        else if (inFs1 && dt2Val == dt1Val) dtAcc.right
-        else  /*inFs1 && dt2Val != dt1Val*/ s"conflicting values found for $f".left
+        if      (!inFs1)                    Result.ok(dtAcc.withField(f, dt2Val))
+        else if (inFs1 && dt2Val == dt1Val) Result.ok(dtAcc)
+        else  /*inFs1 && dt2Val != dt1Val*/ Result.fail(s"conflicting values found for $f")
       })
   }
 
@@ -174,7 +163,7 @@ object InputParsers extends Parsers {
     Parser(in => if (in.atEnd) Success((), in) else Failure("not at EOF", in))
 
   /**
-    * List of `PatternParser`s
+    * Methods for producing `PatternParser`s for a specific time zone.
     *
     * The important one is `pattern`: this parses the whole file pattern
     * and returns the complete file name parser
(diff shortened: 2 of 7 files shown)


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -113,7 +99,7 @@ object InputParsers extends Parsers {
     * Parses a file name and returns the directories which the file should be
     * placed in.
     */
-  type FileNameParser = Parser[List[String]]
+  type FileNameParser = Parser[(List[String], DateTime)]
```
vineethvarghese: Documentation above doesn't talk about `DateTime`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
```diff
@@ -19,42 +19,62 @@ import java.io.File
...
+/** A control file, not to be loaded into HDFS, with timestamp */
+case class ControlFileTimestamped(file: File, val timestamp: DateTime) {
+  def toControlFile = ControlFile(file)
+}
+
+/** A file to be uploaded, with associated business date directory */
 case class DataFile(file: String, parsedDate: String)
 
+/** A file to be uploaded, with associated business date directory and date-time */
+case class DataFileTimestamped(file: String, parsedDate: String, val timestamp: DateTime) {
```
vineethvarghese: Can't we have something like:- 

```scala
case class Timestamped[A](val value: A, timestamp: DateTime) 
```

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
```diff
@@ -19,42 +19,62 @@ import java.io.File
...
+  def toDataFile = DataFile(file, parsedDate)
+}
+
 /** Files found by [[Input.findFiles]] */
 case class InputFiles(controlFiles: List[ControlFile], dataFiles: List[DataFile])
 
+case class InputFilesTimestamped(controlFiles: List[ControlFileTimestamped], dataFiles: List[DataFileTimestamped]) {
```
vineethvarghese: So, are all the files timestamped i.e. both the control files and the data files?

vineethvarghese:
@scurvabilly Can you also take a look at this please? So, that you can slowly start contributing in this space.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
scurvabilly: Also the `val timestamp` seems superfluous.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
```diff
@@ -19,42 +19,62 @@ import java.io.File
...
+/** A file to be uploaded, with associated business date directory */
 case class DataFile(file: String, parsedDate: String)
 
+/** A file to be uploaded, with associated business date directory and date-time */
+case class DataFileTimestamped(file: String, parsedDate: String, val timestamp: DateTime) {
+  def toDataFile = DataFile(file, parsedDate)
+}
+
 /** Files found by [[Input.findFiles]] */
 case class InputFiles(controlFiles: List[ControlFile], dataFiles: List[DataFile])
```
scurvabilly: Could InputFiles be represented with `Map[ControlFile, List[DataFile]]`?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
```diff
@@ -19,42 +19,62 @@ import java.io.File
...
+}
+
 /** Factory for `Input`s */
 object Input {
 
   /**
-    * Find files from a directory matching a given file pattern.
+    * Find files from a directory matching a given file pattern, with `DateTime`s returned.
     *
     * See the description of the file pattern at [[au.cba.com.omnia.meastro.core.task.Upload]]
```
scurvabilly: Just noticed a typo on 'meastro'.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
scurvabilly: Might want to rename `parsedDate` to something like `businessDateDir` or `destDir`. It made it a bit confusing "why do we store the parsed date alongside the datetime?".

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -50,53 +51,38 @@ case class Match(dirs: List[String]) extends MatchResult
...
-
+  /** convert an option to a Result - expected to be incorporated into omnitool. */
+  def fromOption[A](option: Option[A], msg: String): Result[A] =
+    option match {
+      case Some(a) => Result.ok(a)
+      case None    => Result.fail(msg)
   }
 
   /**
     * Pass in table name and file pattern and get back a matching function on file names.
```
vineethvarghese: This documentation needs to change.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -50,53 +51,38 @@ case class Match(dirs: List[String]) extends MatchResult
...
     *
     * Both the pattern parsing and the matching functions can fail with an error message.
```
vineethvarghese: Do we need to say this? `Result` conveys this meaning anyway.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -281,9 +268,9 @@ object InputParsers extends Parsers {
       * We try to ensure that the parser does not allow joda-time to parse
       * a negative year.
       */
-    def timestamp(pattern: String): MayFail[PartialParser] = for {
-      formatter <- MayFail.safe(DateTimeFormat.forPattern(pattern))
-      fields    <- MayFail.fromOption(DateFormatInfo.fields(formatter),
+    def timestamp(timeZone: DateTimeZone)(pattern: String): Result[PartialParser] = for {
```
vineethvarghese: Do we need currying here?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -297,25 +284,22 @@ object InputParsers extends Parsers {
...
+          val parseRes = Result.safe(formatter.parseInto(mutDateTime, underlying, pos))
           val dateTime = mutDateTime.toDateTime
-          parseRes match {
-            case \/-(newPos) if newPos >= 0 => {
-              Success(dateTime, new CharSequenceReader(underlying, newPos))
-            }
-            case \/-(failPosCompl) => {
-              val failPos = ~failPosCompl
+          parseRes.foldMessage(
+            { newPosOrFailPos => if (newPosOrFailPos >= 0) {  // have newPos
```
vineethvarghese: Do we usually format like :- 

```scala
parseRes.foldMessage({ newPosOrFailPos => 
  if (newPosOrFailPos >= 0) {
  else {
  }
},  { msg => Error(msg, in) })
```

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/UploadExecutionSpec.scala
```diff
@@ -36,6 +38,11 @@ Upload execution properties
   can return the files matching the file pattern          $matching
   can upload the files and return upload info             $uploadFile
 
+  won't upload files with hours with no time zone         $withHoursNoTZ
```
jonjuscorpion666: @rowandavies , does that mean we can't have source files without time-zone in name in landing directory anymore ?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
scurvabilly: 'the directories which the file should be placed in' implies to me that the file will be placed in each of these directories, where instead the list represents the path parts of the destination directory, right?

I think it'd also be helpful to explain that the directory parts are actually date parts (Year/Month/Day/...) parsed from the input filename, as this wasn't so obvious to me until I'd dug through the code.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -137,7 +123,7 @@ object InputParsers extends Parsers {
       val combinedParser = for {
```
scurvabilly: On '119' earlier, this looks like a mistake in the documentation:

"It is a parse error (not failure) if the date time fields are consistent."

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -149,22 +135,23 @@ object InputParsers extends Parsers {
       * return a new date time containing the union of all fields. It is an
       * error if the date times have different values for any field
       */
-    def combineDateTimes(dt1: DateTime, fs1: List[DateTimeFieldType],
-      dt2: DateTime, fs2: List[DateTimeFieldType]): MayFail[DateTime] =
+    def combineDateTimes(
```
scurvabilly: May be helpful to clarify that we expect to be parsing a single date time representation and this function is combining that date's parts.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -215,7 +202,7 @@ object InputParsers extends Parsers {
     val timestamp: PatternParser = for {
```
scurvabilly: On line 192 above (I can't seem to comment on it directly), we emit a value `tableName` when the `{table}` literal is encountered. It's not clear to me why we're doing this and how we're using this emitted value later on though. Especially given the parser result type is `type FileNameParser = Parser[(List[String], DateTime)]`, which doesn't allow us to represent this value in any case.

Some documentation behind the reason why we need to parse the literal `{table}` would help here - `UploadExecution` seems like it might be the best place for that.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
```diff
@@ -201,24 +209,64 @@ trait UploadExecution {
...
+    * [[findSourcesTimestamped]] to include hours and/or to return timestamps.
+    *
+    * @param config The upload configuration: [[UploadConfig]].
+    *
+    * @return List of source files in the local ingestion path
+    */
+  def findSources(config: UploadConfig): Execution[List[DataFile]] = for {
+    dataFiles <- UploadEx.matcher(config, DateTimeZone.UTC)
+    _         <- Execution.guard(
+                   dataFiles.forall(_.parsedDate.split(File.separator).length <= 3),
```
scurvabilly: It's a bit mysterious where `<= 3` comes from here. I think we could look up the index using `DateTimeFieldType.hourOfDay` or otherwise add a comment to clarify where `3` is coming from.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
```diff
@@ -201,24 +209,64 @@ trait UploadExecution {
...
     } yield(header, trailer)
   }
 
   /**
+    * Find source files in the local ingestion path without hours in timestamps
+    *
+    * See the description of the files we upload in [[upload]]. Timestamps are
+    * validated as dates, but may not include hours, since their validity
+    * varies during daylight savings transitions.  Use [[findSourcesUTC]] or
+    * [[findSourcesTimestamped]] to include hours and/or to return timestamps.
```
scurvabilly: Can we even be sure of validity when we know the timezone of the input file?

Take AEST: We will always receive a duplicate file for 'April 3 2016 2am'. Even though we know the timezone, we still cannot determine which is which.

To truly solve this, isn't our only option to enforce that the source system to provide files in UTC time?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
```diff
@@ -145,7 +147,13 @@ trait UploadExecution {
     * @return List of copied hdfs files.
     */
   def upload(config: UploadConfig): Execution[UploadInfo] =
-    UploadEx.matcher(config).flatMap(UploadEx.uploader(config, _))
+    findSources(config).flatMap(uploadSources(config, _))
+
+  def uploadTimestamped(config: UploadConfig, timeZone: DateTimeZone): Execution[UploadInfo] =
```
scurvabilly: I think we should comment these functions in a similar way to `findSources...`.

scurvabilly:
Finished with my review, looking good 👍. Most comments are just around documentation.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Addressed:  

```
Parses a file name and returns a list of the path parts to the destination
    * directory, (e.g., for daily files: YYYY/MM/DD) along with the
    * corresponding `DateTime`. Unused components to the `DateTime` default to
    * the least valid value (E.g., for yearly files: January 1st 00:00).
```

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
rowandavies: Yes, they have to be, since both are specific to a particular hour, day, month or year.  In the code, both control filenames and data filenames match `filePattern` in `findFiles` below, including the timestamp, and the matches are then partitioned based on `controlPattern`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
rowandavies: @vineethvarghese I'm aiming to minimise the client code changes when moving from the earlier `DataFile` to `DataFileTimestamped`, and this would require changes to the (by far most common) case when just the `file` is accessed.  

@scurvabilly Ah, good catch on the `val`.  A remnant of something earlier.

@scurvabilly Basically for backward compatibility: `parsedDate` is used in current client code.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
rowandavies: The exact relationship varies between loads, but we could consider supporting some common patterns.  I've just created:  https://github.com/CommBank/maestro/issues/486

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/Input.scala
rowandavies: Fixed.  And fixed one other similar issue found via grep.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: I guess the point in the original code was that there are two distinct phases (parsing and matching) that can fail, and these are different kinds of failures, unlike the usual case where there is just one.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Added `timezone,`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: No, it was leftover from something earlier.  Fixed.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Yes, but in this case then the `(` and `{` opened on the first line should be closed on the same line.  But, this doesn't fit here.  So they need to be on separate lines to match the way they are closed.

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/UploadExecutionSpec.scala
rowandavies: If the timestamps include an hour, then it's not possible to validate whether it is a valid hour without knowing whether it is UTC or Sydney time or whatever.  One hour a year is valid in UTC but not Sydney time.  Generally, in the common case that you want to allow that hour you should use `findSourcesUTC`.

In the even more common case that it's not an intra-day load, just `findSources` still works.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
rowandavies: Agreed, will do.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Ah - I changed `consistent` -> `inconsistent`.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Added to the comment above: 

`since parsers should always yield only a single date time, even if some or all components are repeated.`

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/UploadExecutionSpec.scala
jonjuscorpion666: For projects like site catalyst , we have files with hour and no time zone .  In such case we wont be able upgrade maestro until source add Time zone in file . Correct me if my understanding is wrong .

rowandavies: In such cases you should swap to `findSourcesUTC`, since you don't want to reject the hours (once a year) that don't exist in Sydney time unless you're sure that the time really is Sydney time.

It is deliberate that you need to make this change, since you *are* changing the meaning of your program if you decide to accept these hours, and I think it's appropriate that your tests fail until you go to look at your program and make this decision appropriately.  Adding `UTC` shouldn't be considered an essential code change though - there's no need to have a PR review for just this change.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
rowandavies: DateTimeFieldType.hourOfDay won't work that well because all we can do is check whether the hour is something different from the default of 0.  I posted something on gitter around this, and we should discuss tomorrow - the best solution may be to have the Timestamped versions include the list of date-time fields that were actually parsed.  This could also be useful in things like the load templates.

For now, I've documented and used a constant val to make this clearer:

```scala
    hourPos   = 4   // Hours would be in the 4th component of the parsedDate.
    _         <- Execution.guard(
                   dataFiles.forall(_.parsedDate.split(File.separator).length < hourPos),
                   s"""|Timestamps with hours are deprecated with upload and findSources, instead use
                      | uploadTimestamped, uploadUTC, findSourcesTimestamped or findSourcesUTC.
                      | Timestamps are: ${dataFiles.map(_.parsedDate)}
                   """.stripMargin
                 )
```

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
rowandavies: I've put the following and similar for the UTC variant:

```scala
  /** A variant of [[upload]] that additionally takes a [[timeZone]] in order to
    * correctly support and validate hours, minutes and seconds in timestamps.
    *
    * @param config The upload configuration: [[UploadConfig]].
    * @param timeZone The time zone for timestamps in file name.
    *
    * @return List of copied hdfs files.
    */
```

rowandavies: I also changed the comment for `upload` to say that hours shouldn't be included, and also:

```scala
    *    - `upload` only supports certain date time fields:
    *      - year (y),
    *      - month of year (M),
    *      - day of month (d)
    *    - [[uploadTimestamped]] and [[uploadUTC]] additionally support:
    *      - hour of day (H),
    *      - minute of hour (m), and
    *      - second of minute (s).
```

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -325,9 +330,9 @@ object InputParsers extends Parsers {
             val field = uploadFields(i)
             val value = dateTime.property(field).getField.get(dateTime.getMillis)
             if (field equals DateTimeFieldType.year) f"$value%04d" else f"$value%02d"
-          })
+               })
```
scurvabilly: Rogue tab snuck in.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/InputParsersSpec.scala
```diff
@@ -49,104 +50,115 @@ failing file pattern parsing
   fail on invalid field values            $expectValidFieldValues
 """
 
+  // Helper methods
+  def forPatternUTC = InputParsers.forPattern(DateTimeZone.UTC) _
```
SamRoberts: Perhaps add a new test with a non-standard timezone? After all, you are meant to be able to vary them now.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
```diff
@@ -119,7 +122,7 @@ trait UploadExecution {
     *
     * The file pattern is a string containing the following elements:
     *  - Literals:
-    *    - Any character other than `{`, `}`, `*`, `?`, or `\` represents itself.
+    *    - Any c ofharacter other than `{`, `}`, `*`, `?`, or `\` represents itself.
```
SamRoberts: typo?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
```diff
@@ -130,11 +133,12 @@ trait UploadExecution {
     *    - `upload` only supports certain date time fields:
     *      - year (y),
     *      - month of year (M),
-    *      - day of month (d),
+    *      - day of month (d)
```
SamRoberts: Make the end of this list consistent with the end of the list just below?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
```diff
@@ -130,11 +133,12 @@ trait UploadExecution {
...
+    *    - [[uploadTimestamped]] and [[uploadUTC]] additionally support:
     *      - hour of day (H),
     *      - minute of hour (m), and
     *      - second of minute (s).
-    *
+    *
```
SamRoberts: accidental whitespace?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
```diff
@@ -309,6 +309,11 @@ object InputParsers extends Parsers {
       case _ => Error(s"PartialParser only works on CharSequenceReaders", in)
     }))
 
+    /** The frequencies that should be inferred for corresponding date-time fields */
+    val frequencies = Vector(
+      Period.years(1), Period.months(1), Period.days(1), Period.hours(1), Period.minutes(1), Period.seconds(1)
```
scurvabilly: I suppose it's best to think of this as 'minimum frequency' (i.e. we don't really have source systems generating files every second :smile:).

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/UploadExecutionSpec.scala
```diff
@@ -120,10 +127,10 @@ Upload execution properties
   def uploadFile = withEnvironment(path(getClass.getResource("/upload-sources").toString)) {
     val root      = s"$dir/user"
     val dirStruct = "normal/mydomain/mytable"
-    val destDir   = "hdfs-root/source/$dirStruct"
+    val destDir   = s"hdfs-root/source/$dirStruct"
     val conf      = UploadConfig(
       s"$root/local-ingest/dataFeed/normal/mydomain", destDir,
-      s"$root/local-archive/$dirStruct", "hdfs-root/archive/$dirStruct", "mytable"
```
SamRoberts: embarrassing!

SamRoberts:
looks good to me

scurvabilly:
Yes looks good.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/upload/InputParsersSpec.scala
rowandavies: This was only tested at the higher level, via `UploadExecutionSpec`.  I added a test here as well.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Done.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
rowandavies: Oops.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/UploadExecution.scala
rowandavies: Dammit, I thought that was my cursor.  Fixed.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/upload/InputParsers.scala
rowandavies: Added `(maximum)`

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/UploadExecutionSpec.scala
rowandavies: It did work fine anyway...  it's been doing this for a long time!

vineethvarghese:
@rowandavies Can you squash and clean master's history?

rowandavies:
I already squashed and pushed to Master so that is fine and has the squashed commit but this branch does not.  I'll force push here too but then this will be an empty PR which I'll close and comment with the commit to link the two.

rowandavies:
Apologies, the commit to master for this is:  https://github.com/CommBank/maestro/commit/e06bc75d418e4e0693524f5290650e998cd6bf36

PR #492, opened 2016-06-10
Adding derived db arguement to maestroConfig

bpbinumon:
Adding derived db variables to the config class.


vineethvarghese:
Why do we need this?

bpbinumon:
@vineethvarghese Multiple projects use the derived db variable in their execution class and this has already been added to SimpleMaestroJob class in  etl.util .

rowandavies:
@bpbinumon Thinking about this some more:  

1. Just because it is in etl.util doesn't mean it is appropriate in maestro, since  etl.util is CBA specific, while maestro is open source.  
2. What projects actually use this, and which ones do we expect in future?  I actually don't know of any other than the sap-business-partner PR we just had.

bpbinumon:
@vineethvarghese @rowandavies Currently within datahub , the below projects use derived db variables.
1)etl.sap-businesspartner
2)etl.pods.ccplus
3)etl.ccs-argus
4)etl.commscore-targetedmarketing

And in future, whenever there is a requirement to store non-raw data , we have to use derived db.

rowandavies:
Okay - I can see that those other projects follow the same convention, but so far as I can see they don't actually use the value that's set in etl.util currently.  So, we can decide now to do things a different way without upsetting any existing projects.

I'm not sure what you mean that we have to use the derived db.  Did someone already make this decision (who?) or are you proposing it now?  It's really critical that we track this kind of decision, at least via https://github.com/CommBank/datahub/issues and likely the ODF too if it affects other parts of Omnia.

I'm also not sure that adding the concept of a derived table to maestro is appropriate - it could be, but someone will need to give a rationale for doing this so that we can discuss it and make a decision.  The approach followed for the existing maestro configuration for "raw" and "staging" being set to CBA specific values from etl.util is actually a bit complicated - the rationale for that was achieving backward compatibility.  While @vineethvarghese and myself reviewed that change in https://github.com/CommBank/util.etl/pull/39 I don't think we intended that the derived db be added to maestro.

For the derived db we could instead just define these values within etl.util directly, and that seems more appropriate to me.

Given that @laurencer suggested talking about derived tables at the ODF, I suggest that we do that.  @bpbinumon Can your squad raise a pull request for this on https://github.com/CommBank/OmniaDesignForum ?

bpbinumon:
@rowandavies sure.
Should I decline this PR for now and raise a new one if we get a confirmation from ODF?

PR #493, opened 2016-06-14
Make HiveTable generic to support partition information supplied externally

vineethMM:


rowandavies:
I've added a commit with my suggestion for achieving backward compatibility via a separate 'GenHiveTable`.  The maestro api still needs to be updated though.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -38,8 +38,9 @@ import au.com.cba.omnia.maestro.scalding.ExecutionOps._
   * Base type for partitioned and unpartitioned Hive tables.
   * ST - Type parameter of [[TypedSink]] and [[TypedPipe]]
   * returned in `sink` and `write` methods respectively.
+  * F - Can be any type given a way to extract `ST` from `F`
   */
-sealed trait HiveTable[T <: ThriftStruct, ST] {
+sealed trait GenHiveTable[T <: ThriftStruct, ST, F] {
```
vineethvarghese: I am not sure GenHiveTable is a good name :)

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -51,35 +52,44 @@ sealed trait HiveTable[T <: ThriftStruct, ST] {
...
+  /** Create [[Execution]] that write the given [[TypedPipe]] by extracting `ST` from it.*/
+  def writeExecution(pipe: TypedPipe[F], append: Boolean = true): Execution[ExecutionCounters]
+    = writeToSink(pipe.map( extract ), append)
+
   /** Creates [[Execution]] that writes to `this.sink` using given [[TypedPipe]].*/
-  def writeExecution(pipe: TypedPipe[T], append: Boolean = true): Execution[ExecutionCounters]
+  def writeToSink(pipe: TypedPipe[ST], append: Boolean = true): Execution[ExecutionCounters]
 }
 
+trait HiveTable[T <: ThriftStruct, ST] extends GenHiveTable[T, ST, T]
```
vineethvarghese: Should `extract` just be `id` in this case?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -51,35 +52,44 @@ sealed trait HiveTable[T <: ThriftStruct, ST] {
...
-) extends HiveTable[A, (B, A)] {
+abstract class AbstractPartitionedHiveTable[A <: ThriftStruct : Manifest, B : Manifest : TupleSetter, C]
+  extends GenHiveTable[A, (B, A), C] {
+
+  def partitionBatchSize: Int
+  def fieldNames: List[String]
 
   /** List of partition column names and type (string by default). */
-  val partitionMetadata    = partition.fieldNames.map(n => (n, "string"))
+  val partitionMetadata    = fieldNames.map(n => (n, "string"))
```
vineethvarghese: `val` depending on a `def`?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
rowandavies: Right - it's following the scala api convention that has `GenSeq`, `GenTraversable`, ... for generalised versions of pre-existing things.   I'd agree it could be more specific about what was generalised.  Maybe `HiveTableWithExtractor`?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
rowandavies: Agreed, this is a better place to do that than below.  I was just sketching a suggestion on top of the existing PR to show how to make this backward compatible.

rowandavies: Actually, I think `extract` shouldn't be included in `GenHiveTable` or `HiveTable` instead it should be added as a method in `AbstractPartitionedHiveTable`.

And, rather than the identity, the code below is:
```scala
   override val extract = (r: A) => (partition.extract(r) -> r)
```

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
rowandavies: Agreed - `fieldNames` and `partitionBatchSize` should probably both be `val`s.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethMM: @rowandavies thanks a lot for making it backward compatible.  It make sense to push `extract` method down to `AbstractPartitionedHiveTable`

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethMM: @vineethvarghese  @rowandavies  Thanks for your Inputs.

@rowandavies  can we remove `GenHiveTable ` without having duplicate code in `AbstractPartitionedHiveTable` ? or am I getting it wrong ?

rowandavies: With `extract` moved down, maybe the name could be `HiveTableGenPartition` or `HiveTableAltPartition`.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -54,32 +55,44 @@ sealed trait HiveTable[T <: ThriftStruct, ST] {
...
+  extends GenHiveTable[A, (B, A), C] {
+
+  val partitionBatchSize: Int
+  val fieldNames: List[String]
+
+  def extract: C => (B, A)
 
   /** List of partition column names and type (string by default). */
-  val partitionMetadata    = partition.fieldNames.map(n => (n, "string"))
+  lazy val partitionMetadata    = fieldNames.map(n => (n, "string"))
```
rowandavies: Say why these are lazy.

rowandavies:
With the final suggestions I just made, this should work well.  👍 

@vineethvarghese may want to have a look too.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethMM: what about `HiveTableGenWriteSupport` ?

rowandavies:
Actually, I'd still like to do one last pass here.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -54,32 +55,44 @@ sealed trait HiveTable[T <: ThriftStruct, ST] {
 
   /** Creates a scalding source to read from the hive table.*/
   def source: Mappable[T]
+
+  /** Create [[Execution]] that write the given [[TypedPipe]] by extracting `ST` from it.*/
+  def writeExecution(pipe: TypedPipe[F], append: Boolean = true): Execution[ExecutionCounters]
```
rowandavies: Can we just leave out `writeExecution` here, and instead introduce it in `HiveTable`?  Then, the `F` parameter shouldn't be needed.  `writeSink` could also be left out here and introduced later in `AbstractPartitionedHiveTable`.

Either way, something still seems more complicated than it needs to be here.  (I expect have more time to look at this tomorrow.)

vineethMM: If we push `writeExecution ` down to HIveTable, will it be possible to support GenHiveTable from `viewHive` functionality ?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -40,7 +40,7 @@ import au.com.cba.omnia.maestro.scalding.ExecutionOps._
   * returned in `sink` and `write` methods respectively.
   * F - Can be any type given a way to extract `ST` from `F`
   */
-sealed trait GenHiveTable[T <: ThriftStruct, ST, F] {
+sealed trait HiveTableSource[T <: ThriftStruct, ST] {
```
rowandavies: drop the ST.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -55,24 +55,22 @@ sealed trait GenHiveTable[T <: ThriftStruct, ST, F] {
...
 }
 
-trait HiveTable[T <: ThriftStruct, ST] extends GenHiveTable[T, ST, T]
+trait HiveTable[T <: ThriftStruct, ST] extends HiveTableSource[T, ST] {
+  /** Create [[Execution]] that write the given [[TypedPipe]] by extracting `ST` from `F`.*/
+  def writeExecution(pipe: TypedPipe[T], append: Boolean = true): Execution[ExecutionCounters]
+}
 
 /** Information need to address/describe a specific partitioned hive table.*/
 abstract class AbstractPartitionedHiveTable[A <: ThriftStruct : Manifest, B : Manifest : TupleSetter, C]
```
rowandavies: drop C, rename the others to T and P.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -38,8 +38,9 @@ import au.com.cba.omnia.maestro.scalding.ExecutionOps._
   * Base type for partitioned and unpartitioned Hive tables.
   * ST - Type parameter of [[TypedSink]] and [[TypedPipe]]
   * returned in `sink` and `write` methods respectively.
+  * F - Can be any type given a way to extract `ST` from `F`
```
rowandavies: Remove F from comment, move ST down.

rowandavies:
Looks good, I think this is the best interface now, just a few final suggestions. :+1:

vineethMM:
@vineethvarghese @rowandavies Can you please have a quick look at this PR

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -54,38 +50,53 @@ sealed trait HiveTable[T <: ThriftStruct, ST] {
...
   /** List of partition column names and type (string by default). */
-  val partitionMetadata    = partition.fieldNames.map(n => (n, "string"))
+  lazy val partitionMetadata    = fieldNames.map(n => (n, "string"))
   //Enforce the Hive partition pattern
-  val hivePartitionPattern = partition.fieldNames.map(_ + "=%s").mkString("/")
-  val partitionGlob        = partition.fieldNames.map(_ => "*").mkString("/")
-  val hdfsPartitionGlob    = s"[^_.]$partitionGlob"
+  lazy val hivePartitionPattern = fieldNames.map(_ + "=%s").mkString("/")
+  lazy val partitionGlob        = fieldNames.map(_ => "*").mkString("/")
+  lazy val hdfsPartitionGlob    = s"[^_.]$partitionGlob"
```
vineethvarghese: Why do we need lazy here?

rowandavies: @vineethvarghese This is one way of avoiding issues with the order that the constructors are run. See: http://stackoverflow.com/questions/29673758/getting-a-null-with-a-val-depending-on-abstract-def-in-a-trait 

Another way is just to make all of these `def`s, which has a small performance cost.

We could also have `fileNames` be a def (as it was in the earlier version you commented on), and then have the policy that it's not implemented in subclasses by a `val` (since if it was, the initialisation would be too late) .  But a `lazy val` seems to be the most recommended solution here.  See also:  http://docs.scala-lang.org/tutorials/FAQ/initialization-order.html

vineethvarghese: @vineethMM @rowandavies I would do :- 
```
scala> abstract class T1(names: List[String])
defined class T1

scala> class T2(names: List[String]) extends T1(names) {
     |   def blah = names.toString
     | }
defined class T2

scala> class T2(names: List[String]) extends T1(names) {
     |   val blah = names.toString
     | }
defined class T2
```

and have an abstract class with constructors or just make them all `def`.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -131,25 +141,47 @@ case class PartitionedHiveTable[A <: ThriftStruct : Manifest, B : Manifest : Tup
...
+
+  override def writeExecution(pipe: TypedPipe[T], append: Boolean = true): Execution[ExecutionCounters] =
+    writeToSink( pipe.map(r => (partition.extract(r) -> r)), append)
+}
+
+/**
+  * Information need to address/describe a specific partitioned hive table partition sink
+  * write support.
+  */
+case class GenericPartitionedHiveTable[T <: ThriftStruct : Manifest, P : Manifest : TupleSetter](
```
vineethvarghese: If you can make `AbstractPartitionedHiveTable` *non-abstract*, do you need this anymore?

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
```diff
@@ -54,38 +50,53 @@ sealed trait HiveTable[T <: ThriftStruct, ST] {
...
-  /** Creates [[Execution]] that writes to `this.sink` using given [[TypedPipe]].*/
+}
+
+/**
+  * Base type for hive table with write support.
+  * ST - Type parameter of [[TypedSink]] and [[TypedPipe]] (Kept for backward compatibility)
+  * returned in `sink` and `write` methods respectively.
+  */
+trait HiveTable[T <: ThriftStruct, ST] extends HiveTableSource[T] {
+  /** Create [[Execution]] that write the given [[TypedPipe]] by extracting `ST` from `F`.*/
```
vineethvarghese: What is `F` here?

vineethvarghese:
Done with my review.

Version also need to be bump and we should do a fast forward merge please. No merge commits.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethMM: @vineethvarghese  That make sense, will do that change

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
rowandavies: Agreed, a non-abstract class avoids all these OO-complications with constructor invocation order and is a more "functional programming" design.  Then we just have `val` for all these.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
rowandavies: I think we don't, unless we need `GenericPartionedHiveTable` to be a case class.  We can just add all the constructor arguments as `val` constructor arguments to `AbstractPartitionedHiveTable` and rename that as `GenericPartionedHiveTable` and have `PartionedHiveTable` extend that.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/hive/HiveTable.scala
vineethMM: Thats a out dated comment :)

rowandavies:
That commit is closer - now rename `AbstractPartionedHiveTable` to `GenericPartionedHiveTable` and make it non-abstract and add the rest of the constructor arguments from `GenericPartionedHiveTable` as val arguments.  Unless that doesn't work for some reason?

vineethMM:
@rowandavies if we make `AbstractPartionedHiveTable ` a  non-abstract class we end up in having override modifiers for all  members in `PartitionedHiveTable` or  we have to rename and pass them into `AbstractPartionedHiveTable` class constructor

version.sbt
```diff
@@ -12,7 +12,7 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-version in ThisBuild := "2.20.0"
+version in ThisBuild := "2.20.1"
```
vineethvarghese: This is a 2.21 really not a bug fix version.

vineethMM: @vineethvarghese Thanks for pointing out, will update it

PR #498, opened 2016-08-29
Add Encode for Short datatype

sibisasidharan:


version.sbt
```diff
@@ -12,7 +12,7 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-version in ThisBuild := "2.22.0"
+version in ThisBuild := "2.22.1"
```
rowandavies: Make it 2.23.0 since this adds a new feature, even if it's small.

rowandavies:
Looks good to me.  Just change the version.  :+1:

Issue #497, opened 2016-08-18
Usage of default AET timezone


SajeevAr:
This is related to issue #481. Opening a new issue to track this separately.

Issue: The current timezone addition provides the ability to specify timezone associated with a file during upload and hence fixes the issues in upload step. However, apart from upload, the default AET timezone was assumed in the following instances as well, which may lead to job failures.
1. Methods `byYear`, `byMonth`, `byDay`, `byHour`, `byDate` etc. in `Partition.scala` will parse the timestamp in partition field based on a given format. So `viewHive` step will fail if the timestamp is invalid in default timezone.
2. `isDate` method in `Validator.scala` will fail if timestamp in validator field is invalid in default timezone.
3. `now` method in `TimeSource.scala` also use the default timezone to generate `TimeSource`. It doesn't cause a job failure, but it restricts from creating `TimeSource` in any specific timezone.

To fix these issues, i think we can add the capability to specify desired timezone for above scenarios.

rowandavies:
For `byYear`, `byMonth` and `byDate`, the set of valid dates is the same for every time zone, so it can stay the way it is, or always use UTC via `DateTimeFormat.forPattern(dateFormat).withZoneUTC()` or we can have a timeZone that defaults to UTC (similar to the previous timezone fix #483).

For `byHour` we could do the same, but then there's the potential for a column that is Sydney-time to be treated as UTC by not being updated.  This may be okay here though, since this isn't intended to be where we do validation of the data, and the parsed date-time is just used to extract the fields, which always exactly match the data.  I.e., we never do time zone adjustment here, this isn't the right place to do that.

`isDate` is the right place to do validation, so we should have `isDateUTC` and `isDateTZ` and require these to be used if an hour is included in the format.  Alas, I don't see a nice way to check for hours in patterns via http://joda-time.sourceforge.net/apidocs/org/joda/time/format/DateTimeFormatter.html

For the `TimeSource` `now` method, the date will be different, so we should probably always specify the timezone here - so I'm thinking there will be a new timezone argument with no default, and `nowUTC` and perhaps `nowSydney` for convenience.

Feedback welcome.

rowandavies:
cc @AnnTomie @sibisasidharan @josephshobin

rowandavies:
After discussing with @SajeevAr for the first part, it's not true that "it can stay the way it is" because the  format may include hours even for `byYear`, `byMonth` and `byDate`.

rowandavies:
[...]

PR #507, opened 2016-11-04
Fix byHour daylight savings transition and other time zone dependencies.

rowandavies:
Addresses #497.

This includes some small API changes, just to force existing client code with unintended time zone dependencies to be revisited, similar to #483:

- `Check.isDate` now throws an exception if the pattern is time-sensitive and no time zone is specified.  Replace with `isDateUTC` or `isDateSydney` or explicitly give the intended time zone.

- `TimeSource.now` now requires an additional argument for the time zone.  For convenience, use `nowUTC` or `nowSydney`.

### DIFF
--- a/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
+++ b/maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
@@ -49,14 +49,14 @@ object Partition {
   /** Partition by year, month, day for a given dateFormat. */
   def byDate[A](date: Field[A, String], dateFormat: String = "y-M-d"): Partition[A, (String, String, String)] =
     Partition(List("year", "month", "day"), v => {
-      val dt = DateTimeFormat.forPattern(dateFormat).parseDateTime(date.get(v))
+      val dt = DateTimeFormat.forPattern(dateFormat).parseLocalDateTime(date.get(v))
       (dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d")
     }, "%s/%s/%s")
 
   /** Partition by year, month, day, hour for a given dateFormat. */
   def byHour[A](date: Field[A, String], dateFormat: String = "y-M-d-H"): Partition[A, (String, String, String, String)] =
     Partition(List("year", "month", "day", "hour"), v => {
-      val dt = DateTimeFormat.forPattern(dateFormat).parseDateTime(date.get(v))
+      val dt = DateTimeFormat.forPattern(dateFormat).parseLocalDateTime(date.get(v))
       (dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d", f"${dt.getHourOfDay}%02d")
     }, "%s/%s/%s/%s")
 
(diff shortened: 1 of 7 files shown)


rowandavies:
TODO: update version.

rowandavies:
@vineethvarghese @andynygard @SajeevAr @AnnTomie @sibisasidharan @josephshobin

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/validate/Validator.scala
```diff
@@ -22,7 +22,7 @@ import scalaz._, Scalaz._
 import org.apache.commons.validator.routines._
 import org.apache.commons.validator.GenericValidator
 
-import org.joda.time.DateTime
+import org.joda.time.{DateTime, DateTimeZone,Duration}
```
andynygard: Spacing here

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/validate/CheckSpec.scala
```diff
@@ -26,10 +30,17 @@ object CheckSpec extends Spec { def is = s2"""
...
-  Check.nonempty              $nonempty
-  Check.nonempty failure      $nonemptyFailure
+  Check.oneOf                        $oneOf
+  Check.oneOf failure                $oneOfFailure
+  Check.formatIncludesTimeTrue       $formatIncludesTimeTrue
+  Check.formatIncludesTimeFalse      $formatIncludesTimeFalse
+  Check.isDateNoHour                 $isDateNoHour
+  Check.isDateNoTZ                   $isDateNoTZ
+  Check.isDateDaylightSavingsSkip    $isDateDaylightSavingsSkip
+  Check.isDateNoDaylightSavingsSkip  $isDateDaylightSavingsSkip
```
SajeevAr: Should we call `isDateNoDaylightSavingsSkip` instead ?

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
```diff
@@ -377,8 +384,12 @@ View execution properties
   def tableBySecond(database: String) =
     HiveTable(database, "by_second", bySecond, None)
 
+  def tableByHour(database: String) =
```
SajeevAr: Do we need this method ?

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/validate/Validator.scala
SajeevAr: Space before `Duration`

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/time/TimeSource.scala
```diff
@@ -42,9 +42,15 @@ object TimeSource {
     predetermined(f.print(time))
   }
 
-  /** Use the current time as the time source */
-  def now(format: String = "yyyy-MM-dd"): TimeSource =
-    predetermined(new DateTime, format)
+  /** Use the current time as the time source, using a specified time zone. */
+  def now(format: String = "yyyy-MM-dd", tz: DateTimeZone): TimeSource =
+    predetermined(new DateTime(tz), format)
```
andynygard: Forcing this to be explicit seems good.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
```diff
@@ -49,14 +49,14 @@ object Partition {
...
-      val dt = DateTimeFormat.forPattern(dateFormat).parseDateTime(date.get(v))
+      val dt = DateTimeFormat.forPattern(dateFormat).parseLocalDateTime(date.get(v))
       (dt.getYear.toString, f"${dt.getMonthOfYear}%02d", f"${dt.getDayOfMonth}%02d")
     }, "%s/%s/%s")
 
   /** Partition by year, month, day, hour for a given dateFormat. */
   def byHour[A](date: Field[A, String], dateFormat: String = "y-M-d-H"): Partition[A, (String, String, String, String)] =
     Partition(List("year", "month", "day", "hour"), v => {
-      val dt = DateTimeFormat.forPattern(dateFormat).parseDateTime(date.get(v))
+      val dt = DateTimeFormat.forPattern(dateFormat).parseLocalDateTime(date.get(v))
```
andynygard: I had a think through the scenarios of old style vs new style, as I was concerned that we may inadvertently change the values of the partitions we generate (e.g. are we sure that the partitions won't shift by 11 hours after this change).

Thinking it through this doesn't seem to be the case, `toString` should always give the same numeric Y/M/D/H values as in the input string. And using `LocalDateTime` will eliminate the non-existent hour problem 👍

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/validate/Validator.scala
```diff
@@ -65,17 +65,36 @@ object Check {
...
-    lazy val formatter = DateTimeFormat.forPattern(pattern)
+  def formatIncludesTime(formatter: DateTimeFormatter) = {
+    val someTime = new DateTime("2016-11-03T03:30:00.000-04:00")  // A time used to probe formats sensitivity to hours.
+
+    // We add a duration of 01:01:01.001 twice and see if this changes the result both times.
+    val step = Duration.standardHours(1).plus(Duration.standardMinutes(1)).plus(new Duration(1001))
+    def formattedPlusSteps(numSteps: Int) = formatter.print(someTime.withDurationAdded(step, numSteps))
+
+    // Check whether the string changes twice in the next two hours.
+    (formattedPlusSteps(0) != formattedPlusSteps(1)) && (formattedPlusSteps(1) != formattedPlusSteps(2))
```
andynygard: Maybe we could simplify this by formatting `someTime` to string, then parsing it back into a `DateTime` using the same formatter. We can then check, does this `DateTime` have a time component (e.g. `dt.withTimeAtStartOfDay() != dt`)?

SajeevAr:
Looks good, otherwise 👍

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/validate/Validator.scala
rowandavies: Fixed.

rowandavies: Fixed.

maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/ViewExecutionSpec.scala
rowandavies: Removed.

maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/partition/Partition.scala
rowandavies: Fixed.

rowandavies: Right.  The concept of timezone doesn't really have a place here, so `LocalDateTime` seems to be what we want.

PR #516, opened 2016-11-28
IO -> Execution

triggerNZ:
Conversion between scalaz.effect.IO and scalding Execution


vineethvarghese:
version bump?

rowandavies:
Maybe add a ~> natural transformation?  (Not essential.)

triggerNZ:
@rowandavies I really like the `~>` idea but think that if we are doing that we should write `~>` instances for all of the things we are converting from (not just `IO`) and have a general

```scala
def from[M[_], V](m: M[V])(implicit nat: M ~> Execution): Execution[V] = ...
```

This feels like a bit of a refactor and probably should be in a separate PR

triggerNZ:
@vineethvarghese I bumped the version. Other than that are you happy for me to merge?

vineethvarghese:
👍

PR #521, opened 2017-01-11
Automatic version update

toddmowen:


toddmowen:
Autowrangler is not very robust. Network dropout to artifactory caused this misbehaviour...

PR #531, opened 2017-05-01
Support jobnotready and jobfailures wrapped in result exception during error recovery

jithinjohn:
This is to Support `JobNotReadyException` and `JobFailureException` wrapped in `ResultException`  returning expected values during error recovery.

### DIFF
--- a/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/scalding/MaestroExecution.scala
+++ b/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/scalding/MaestroExecution.scala
@@ -25,11 +25,17 @@ object MaestroExecution {
   def jobFailure[T](exitCode: Int = -1): Execution[T] =
     Execution.from[T](throw new JobFailureException(exitCode))
 
-  /** Lifts a [[JobNotReady]] or [[JobFailure]] error into a JobStatus value for execution. */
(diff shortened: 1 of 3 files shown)


andynygard:
Looks like autowrangler beat you to the version bump :) It's probably best to rebase master again in order to avoid a messy merge history.

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/scalding/MaestroExecution.scala
```diff
@@ -28,8 +28,10 @@ object MaestroExecution {
   /** Lifts a [[JobNotReady]] or [[JobFailure]] error into a JobStatus value for execution. */
```
andynygard: I think we should document why we would want to handle `ResultException` and how it could arise.

Something like,
```scala
/** Lifts a [[JobNotReady]] or [[JobFailure]] error into a [[JobStatus]] value for execution.
  * 
  * Due to the way that Omnitool embeds [[Result]] failures into [[Execution]], [[ResultException]]
  * is checked (see `resultToFuture` in [[RichExecution]]).
  */
```

jithinjohn: @andynygard  Updated the same. Please have a look

Issue #425, opened 2015-07-09
Support for multiple character set


anandsiva:
Requirement to support UTF character set  so that source can send and load those datasets

jdcba:
I assume the original reason for requiring sources to be ASCII-only is due to lack of mappings from EBCDIC? Or possibly a GDW 1.0 tradition? Maestro has been filtering out any errant non-ASCII characters that might come through. It was also depending on the defaults in Cascading/Scalding.

My approach with PR#532 and PR#533 is to make reader charset explicit in the public API (converts to Java Unicode string) and provide a corresponding Unicode-capable filter.

jdcba:
Hi all, v2.28.0 includes support for custom character sets. The Maestro legacy default (per Scalding/Cascading) remains as-is:

1. Parses UTF-8 only
2. Replaces any non-UTF-8 characters silently
3. Downselects to US-ASCII printables only

If the original source is not UTF-8 or US-ASCII, the new option `LoadConfig.charsetEncoding` can be used to resolve issues 1-2.

The new option `Clean.unicodeDefault` can be used to preserve Unicode characters during the load to resolve issue 3.

PR #532, opened 2017-06-08
Add Clean.unicodeDefault (AIDH-1863)

jdcba:
v2.27.0 adds optional support for Unicode character handling when `Clean`ing `Field`s. Adds missing unit tests.

This extends Maestro in direct analogy to its existing POSIX US-ASCII charset handling: preserves Printable characters only (effectively trims leading/trailing whitespace & filters out control characters).

Leverages Unicode TR#18 Annex C support in Java’s `Pattern`s as the _de facto_ acceptable character policy (see also https://github.com/CommBank/OmniaDesignForum/blob/289dc0331087ac7ee90f6957a0d2c6107c9106fe/features/string-features.md).

This is intended as a stepping stone for #425 because we have an in-flight job requirement.


maestro-core/src/main/scala/au/com/cba/omnia/maestro/core/clean/Clean.scala
```diff
@@ -30,11 +30,23 @@ object Clean {
   def trim: Clean =
     Clean((_, data) => data.trim)
 
+  /** Keep POSIX US-ASCII printables only.
+    * @see https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#posix
```
rowandavies: Nice.  I was just about to ask for a reference on this so I could review quicker.

maestro-core/src/test/scala/au/com/cba/omnia/maestro/core/clean/CleanSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  val allAscii = (0 to 127).map(_.toChar).mkString
+
+  val posixPrintable =
+    " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~"
+
+  val isPosixPrintable = posixPrintable.toSet
+
+  def isUnicodePrintable(c: Char): Boolean =
+    c.toString.matches("(?U)[\\p{Graph}\\p{Blank}&&[^\\p{Cntrl}]]")
```
rowandavies: Can you teach me quickly in person so I'm confident that this pattern tests correctly?  Or else a second pair of eyes would be sufficient.

rowandavies: Doh!  Andy beat me to it.  Never mind.

jdcba: The Scaladoc references http://www.unicode.org/reports/tr18/#Compatibility_Properties which says: `\p{graph} \p{blank} -- \p{cntrl}` and the Scaladoc also references https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#posix which says “\p{Print}	A printable character: [\p{Graph}\p{Blank}&&[^\p{Cntrl}]]” and “when UNICODE_CHARACTER_CLASS flag is specified” and “The UNICODE_CHARACTER_CLASS mode can also be enabled via the embedded flag expression (?U).” Therefore, putting it all together, I came up with line 44.

jdcba:
Thanks guys! Merging...

PR #533, opened 2017-06-08
Add LoadConfig.charsetEncoding (AIDH-1863)

jdcba:
Further progress on #425 (see #532 for context). Am I on the right track? Will we be able to leverage this via ETL.Template??

Note: Also need unit tests for a non-default character set encoding (try `windows-1252`). I do not have time today, Mithun can you do this if is the support we need?


maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/LoadExecution.scala
```diff
@@ -124,27 +123,32 @@ object LoadInfo {
...
-  * @param validator:      The `Validator` to check the thrift struct to ensure it is correct.
+  * @param validator       The `Validator` to check the thrift struct to ensure it is correct.
   *                        Defaults to no validation.
-  * @param errorThreshold: The fraction of rows that can fail during the load step
+  * @param errorThreshold  The fraction of rows that can fail during the load step
   *                        before the overall job should fail. Defaults to 0.05 or 5%.
-  * @param splitSize:      The size of each split in the job that processes the
+  * @param charsetEncoding The Scalding text encoding (Cascading charset) for parsing text files.
+  *                        Defaults to `TextLine.DEFAULT_CHARSET`, which is UTF-8.
+  *                        By default, `Clean.default` will remove non-ASCII characters,
```
rowandavies: I had to read this at least twice.  Maybe say: 

> The default for the `clean` parameter above, `Clean.default`, will remove ...

maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/LoadExecution.scala
```diff
@@ -188,12 +193,18 @@ trait LoadExecution {
   * We may change this without considering backwards compatibility.
   */
 object LoadEx {
+  /** Like `com.twitter.scalding.MultipleTextLineFiles` but supports an explicit `TextSourceScheme.textEncoding`.
+    * @param textEncoding a charset string (upstream default was `cascading.scheme.local.TextLine.DEFAULT_CHARSET`).
+    */
+  case class MultipleTextLineFilesWithTextEncoding(override val textEncoding: String, p: String*)
+    extends FixedPathSource(p: _*) with TextLineScheme
```
rowandavies: Nice.  I like Scala and the Scalding API again.

rowandavies:
Looks right to me.  A test could be nice here.

rowandavies:
Oops.

rowandavies:
I guess this is only to add to the other branch?  Shall I do a final review on that branch later?

jdcba:
Doh! I am used to the simplicity of Stash doing FF merges so that PRs can be merged cleanly. Will adapt this branch to target master in GH now that the other PR has been merged.

jdcba:
Is anyone willing to approve this or is there something else to consider?

jdcba:
Hi all, this PR leaves the Maestro legacy default as-is:

* Parses UTF-8 only,
* Replaces any non-UTF-8 characters silently, then 
* Downselects to US-ASCII printables

This PR adds the ability to choose a custom parser charset and custom cleaning filter. Together, these options can be used to parse non-UTF-8 and to preserve non-US-ASCII printables.

Most people will want these two options to go hand-in-hand. I think we can add a custom charset option to etl.template to configure both options at the same time.

PR #552, opened 2018-10-28
Sqoop export will no-op for an empty export directory

scurvydoggo:
The version of Sqoop included in the CDH 5.13 binaries contains a breaking change where the sqoop export operation will fail if the directory is empty.

This adds a conditional check prior to this operation in order to keep it compatible.


maestro-scalding/src/test/scala/au/com/cba/omnia/maestro/task/SqoopExportExecutionSpec.scala
```diff
@@ -102,9 +104,10 @@ object SqoopExportExecutionSpec
       val config = SqoopExportConfig(
         options(table).sqlQuery(s"select * from $table"), deleteFromTable = true
       )
-      execute(sqoopExport(config, exportDir)) must throwA[RuntimeException](
-        message = "SqoopOptions.getSqlQuery must be empty on Sqoop Export with delete from table"
-      )
+      execute(sqoopExport(config, exportDir)) must beLike {
+        case Failure(ex: RuntimeException) =>
+          ex.getMessage mustEqual "SqoopOptions.getSqlQuery must be empty on Sqoop Export with delete from table"
+      }
```
scurvydoggo: Now that `SqoopEx.sqoopExport` is embedded in a flatMap, seems that errors are caught an embedded in the `Execution`, rather than getting thrown.

Technically this is a breaking change, however it is also our preferred error handling strategy. Also, our users would already be performing error handling at the monad-level, so it's not like this error would get swallowed after this change.

rowandavies: Agreed.

rowandavies:
There is the option of moving this check to Parlour.  I think I'm happy with having it in Maestro though - Parlour is a relatively low-level wrapper around Sqoop, while Maestro has some higher-level conventions.  I think it's consistent with the existing conventions to have "empty Sqoop exports succeed".  

But, we'd have a little more backward compatibility with this in Parlour.  What do you think?

scurvydoggo:
Yeah @rowandavies , that's true, we could move it to parlour. It would centralise it even better.

We'd just have to a `permafrost` dependency, so that it can use `Hdfs.files`. But that seems OK.. sqoop already reads HDFS, so the behaviour is consistent.

rowandavies:
I think the need for the additonal dependency pushes me towards thinking that this belongs in Maestro instead.  But, we do have at least one direct use of Parlour:
   https://stash.odp.cba/projects/OMETL/repos/util.bicd-parlour/browse/src/main/scala/au/com/cba/omnia/etl/bicd/parlour/ExportSqoopConsoleJob.scala

rowandavies:
I'm approving here on the basis that we need to unblock this so we can have a full set of application builds this evening.

scurvydoggo:
@rowandavies sure, we can always re-iterate to a lower level.

aswathysethu:
Sqoop export can be safely skipped in case of empty export , so that the job succeeds.

However, in case of jobs which checks a record count match with teradata and hive , there will be a mismatch as the previous entry in teradata will not be cleared as we are skipping the export.

rowandavies:
@aswathysethu Just to be clear:  this workaround won't cause any issues, right?  I'm not sure I understood the second part fully, but talking with @snehalvidyan he said there won't be issues.

mbalassi:
@aswathysethu Could you specify how the empty export and the 0 row export will be different when counting rows on the Teradata side? Thanks!

aswathysethu:
@rowandavies It will not cause any issues.

aswathysethu:
@mbalassi 
Hive count and Teradata row count match is done within Omnia, to ensure the exported records are same.

In case of empty exports, we will be updating TCF with 0 row count.

PR #553, opened 2018-10-29
Adding env-input-prefix and env-output-prefix for DIL, ENR and ACCESS…

sujith1984:
Adding env-input-prefix and env-output-prefix for DIL, ENR and ACCESS Layer

### DIFF
--- a/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/MaestroConfig.scala
+++ b/maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/MaestroConfig.scala
@@ -69,6 +69,8 @@ case class MaestroConfig(
   lazy val connString      = args("jdbc")
   lazy val username        = args("db-user")
   lazy val password        = Option(System.getenv("DBPASSWORD")).getOrElse("")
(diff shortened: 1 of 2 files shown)


maestro-scalding/src/main/scala/au/com/cba/omnia/maestro/task/MaestroConfig.scala
```diff
@@ -69,6 +69,8 @@ case class MaestroConfig(
   lazy val connString      = args("jdbc")
   lazy val username        = args("db-user")
   lazy val password        = Option(System.getenv("DBPASSWORD")).getOrElse("")
+  lazy val envInputPrefix  = args.optional("env-input-prefix").getOrElse(args("env"))
```
rowandavies: Maybe add a short comment to explain when this would be needed.

jdcba: I think we can’t use “env” as the default because it will be a full environment like “physical.logical” (“datascience.dev”). Instead, I think we can use mandatory args (because envInputPrefix and envOutputPrefix are lazy vals: they are themselves optional and only accessed if needed):
```suggestion
  lazy val envInputPrefix  = args("env-input-prefix")
```

jdcba:
This is part of supporting latest util.etl

